Issue #300, opened 2023-06-01
Make pairfloes use floats for timedeltas


cpaniaguam:
During integration of soit, generated timedeltas ended up as floats. The function signature needs updating accordingly.
https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/a02afb6b06aa3d9d231621cc976220b239ffd95c/src/tracker/tracker.jl#L29-L35

PR #301, opened 2023-06-01
fix: timedeltas as floats

cpaniaguam:
#300


tdivoll (approved):
Looks pretty straightforward to me!

Issue #312, opened 2023-06-21
Update regionprops_table docstring example


cpaniaguam:
- [ ] Add `using IceFloeTracler, Random`
- [ ] Remove `dataframe=true`
- [ ] Use IceFloeTracker.label_components
https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/97cf1ba7f48a9bb5fa4a7b3e178d86acf957a486/src/regionprops.jl#L25-L56

PR #313, opened 2023-06-21
fix: update regionprops_table docstring example

cpaniaguam:
Came across this while prototyping things with HDF5. #312


tdivoll (approved):
Good bug find!

Issue #314, opened 2023-06-27
Include orientation in default props for regionprops_table


cpaniaguam:
https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/18819f72e1a38b708f28ec7573b2659819804aab/src/regionprops.jl#L64-L72

PR #315, opened 2023-07-05
feat: add orientation to default props

cpaniaguam:
#314


tdivoll (approved):
This looks good to me!

PR #328, opened 2023-08-16
feat: add getlatton

cpaniaguam:


src/IceFloeTracker.jl
```diff
@@ -65,13 +65,14 @@ include("branch.jl")
...
 
 const sk_measure = PyNULL()
+const getlatlon = PyNULL()
 
 function __init__()
     copy!(sk_measure, pyimport_conda("skimage.measure", "scikit-image=0.20.0"))
     pyimport_conda("pyproj", "pyproj=3.6.0")
     pyimport_conda("rasterio", "rasterio=1.3.7")
-    pyimport_conda("requests", "requests=2.31.0")
-    pyimport_conda("skyfield", "skyfield=1.45.0")
```
tdivoll, line 74: Will we still need these if a user wanted to fetchdata without the pipeline and container?

tdivoll: I forgot SOIT was integrated and now the script is only in the pipeline package. So if a user was going to use SOIT they will need to call it from the pipeline package, which works fine with the `icefloetracker-fetchdata` docker container, both locally and on hpc.

tdivoll (approved):
Looks good to me!

src/IceFloeTracker.jl
cpaniaguam: So this is fine. right?

PR #331, opened 2023-08-19
Feat add getiftversion

cpaniaguam:
This is an idea that occurred to me overnight. I think it might work!


tdivoll (approved):
This looks good! I'm not sure why the CI would fail? I tested it and the function works as expected: 
image

cpaniaguam:
> This looks good! I'm not sure why the CI would fail? I tested it and the function works as expected: image

It's a weird error. When the tests are running `Pkg.TOML.parsefile("Project.toml")` doesn't contain the "version" key. It might be a Julia bug. We could handle the error so CI completes, or just hard-code the version number and remember to update it every time changes are made 😖

cpaniaguam:
> > This looks good! I'm not sure why the CI would fail? I tested it and the function works as expected: image
> 
> It's a weird error. When the tests are running `Pkg.TOML.parsefile("Project.toml")` doesn't contain the "version" key. It might be a Julia bug. We could handle the error so CI completes, or just hard-code the version number and remember to update it every time changes are made 😖

I just recalled that there is another Project.toml file in the test folder. I guess I'll add the version number there and see it that fixes it.

cpaniaguam:
Seems to have worked!

test/test-misc.jl
```diff
@@ -0,0 +1,3 @@
+println("------------------------------------------------
+----------------- Misc. Tests ------------------")
+@test getiftversion() == "0.2.0"
```
tdivoll, line 3:
```suggestion
@test getiftversion() == "0.2.0"
@test getiftversion() != "unknown"
```

Maybe adding this will satisfy codecov? Codecov is so demanding :)

tdivoll (approved):

test/test-misc.jl
cpaniaguam: I think it's the try catch block inside the `__init__` function that is not "getting tested". And I don't think there is a way to run that block unless I undo the change in `test/Project.toml`!

cpaniaguam:
@JuliaRegistrator register

PR #332, opened 2023-08-20
feat: hardcode pkg version number

cpaniaguam:
This approach is working well with IFTPipeline!

### DIFF
--- a/src/IceFloeTracker.jl
+++ b/src/IceFloeTracker.jl
@@ -39,7 +39,7 @@ export readdlm,
     matchcorr,
     centered,
     imrotate,
-    getiftversion
+    IFTVERSION
 
 # For IFTPipeline
 using HDF5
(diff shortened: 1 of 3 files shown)


tdivoll (commented):
I was testing with `Pkg.project().version` instead of `VersionNumber(0,2,0)`. It seems to return the version fine and updating the test worked as well. However, it tries to activate a Julia REPL and IFT when I try to commit the changes.

src/IceFloeTracker.jl
```diff
@@ -67,14 +67,9 @@ include("special_strels.jl")
 
 const sk_measure = PyNULL()
 const getlatlon = PyNULL()
-const _version = []
+const IFTVERSION = VersionNumber(0,2,0)
```
tdivoll, line 70:
```suggestion
const IFTVERSION = Pkg.project().version
```

test/test-misc.jl
```diff
@@ -1,3 +1,3 @@
 println("------------------------------------------------
 ----------------- Misc. Tests ------------------")
-@test getiftversion() == "0.2.0"
+@test IFTVERSION == VersionNumber(0,2,0)
```
tdivoll, line 3:
```suggestion
@test IFTVERSION == Pkg.project().version
```

cpaniaguam:
> I was testing with `Pkg.project().version` instead of `VersionNumber(0,2,0)`. It seems to return the version fine and updating the test worked as well. However, it tries to activate a Julia REPL and IFT when I try to commit the changes.

I'll test this within IFTPipeline. Thanks!

cpaniaguam:
> > I was testing with `Pkg.project().version` instead of `VersionNumber(0,2,0)`. It seems to return the version fine and updating the test worked as well. However, it tries to activate a Julia REPL and IFT when I try to commit the changes.
> 
> I'll test this within IFTPipeline. Thanks!

When used within IFTPipeline it returns IFTPipeline's version :/. Going back to testing with hard-coding the version number.

cpaniaguam:
@tdivoll I think hard coding the version number is probably the way to go.

cpaniaguam:
> @tdivoll I think hard coding the version number is probably the way to go.

@tdivoll Things working well within the container.

tdivoll (approved):
Sounds good to me, it is a minor task to update when we update version numbers.

PR #335, opened 2023-08-23
update: track notebook

cpaniaguam:
To address changes in format of input params to  the `pairfloes` function.


tdivoll (approved):
This looks good to me. Was the issue the need for float values or the order of goodness and comp inputs?

Issue #343, opened 2023-09-22
- [ ] Move process_soit from IFTPipeline to IFT to get floe datetimes. Export


cpaniaguam:
- [ ] Move process_soit from IFTPipeline to IFT to get floe datetimes. Export

_Originally posted by @cpaniaguam in https://github.com/WilhelmusLab/IceFloeTracker.jl/issues/339#issuecomment-1731661211_

PR #344, opened 2023-09-22
feat: process_soit for floe tagging

cpaniaguam:
#343


src/tracker/tracker-funcs.jl
```diff
@@ -497,3 +497,48 @@ function addfloemasks!(props, imgs)
...
+The SOIT output file is a csv file with columns :Date, :sat1, :sat2, ... :satn, where each row is a pass time for the satellites listed in the columns. The resulting DataFrame has columns `:sat` and `:pass_time` sorted by :pass_time.
+
+# Arguments
+- `passtimesdir::String`: The directory containing the SOIT output file
+"""
+function process_soit(passtimesdir::String)
+    # check there is a file at passtimesdir starting with "passtimes_lat" with extension .csv
+    pth = filter(x -> occursin(r"^passtimes_lat.*\.csv$", x), readdir(passtimesdir))
+    isempty(pth) && error("No csv file found at $passtimesdir starting with 'passtimes_lat'")
+    length(pth) > 1 && error("More than one csv file found at $passtimesdir starting with 'passtimes_lat'")
```
tdivoll, line 515: Should we handle the multiple soit files during param cycling? The results structure looks different when expanding parameters: 
image

tdivoll (commented):
The code looks good, I was just wondering if we should try to handle the multiple files differently, maybe using the date that is passed to the function?

src/tracker/tracker-funcs.jl
cpaniaguam: @tdivoll This is now handled by cycl. I think this is okay to merge?

cpaniaguam:
These features will be invoked from the pipeline repo so they don't need to be moved over here.

PR #366, opened 2024-04-02
chore: update version nums

cpaniaguam:
Two more places needed updating.


tdivoll (commented):
This looks good. Is there a way to grab the version number from the `Project.toml` directly so we don't have to update it in so many places.

cpaniaguam:
I was able to grab the version number from the Project.toml to define `IFTVERSION`. The updated test also checks this.

tdivoll (approved):
This looks great, and simplifies the process!

PR #368, opened 2024-04-03
test: first attemp to set up trigger workflows to build containers

tdivoll:
This PR adds trigger actions that will set off the Docker build actions at [ice-floe-tracker-pipeline](https://github.com/WilhelmusLab/ice-floe-tracker-pipeline).


tdivoll:
As far as I can tell, the new actions added here will trigger the container builds in the pipeline repo when this PR is merged. This test simulates a scenario where updates are made to this repo and merged into main.

tdivoll:
@cpaniaguam I actually just realized that we don't need to rebuild the fetchdata container; any changes to that would only happen in the pipeline repo, which would trigger the workflow there. We only need the Julia changes to trigger a new build to the Julia container.

README.md
```diff
@@ -79,3 +79,6 @@ git push
     - Click on `Draft a new release`
     - Choose the new tag you created
     - Click on `Generate release notes` or add a custom description
+6. Head over the `ice-floe-tracker-pipeline` [repo](https://github.com/WilhelmusLab/ice-floe-tracker-pipeline/blob/main/Project.toml) and update the `[compat]` section of the `Project.toml` where `IceFloeTracker` is a dependency.
+
+**Note:** After the PR for the release is merged, a trigger workflow will force a rebuild of Docker container used in the [pipeline](https://github.com/WilhelmusLab/ice-floe-tracker-pipeline).
```
cpaniaguam, line 84: I am thinking that adding a special tag could be useful. Maybe create a separate workflow to be triggered by the triggering one on this repo? I am not sure how the tags that appear on docker hub are created.

cpaniaguam (approved):

PR #372, opened 2024-04-05
fix: update trigger-docker-julia.yml to trigger Docker build on release

cpaniaguam:


tdivoll (approved):
Nice! This makes sense. I guess the action just checks for changes in the version number?

tdivoll:
We'll have to make a new minor release to test it.

cpaniaguam:
> We'll have to make a new minor release to test it.

I guess so. It could be a patch also. I can add the label `[published]` to the current release to see if that'll trigger the other action as well.

PR #379, opened 2024-04-08
378 add compathelper

cpaniaguam:
#378

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -16,7 +16,7 @@ jobs:
         run: which julia
         continue-on-error: true
       - name: Install Julia, but only if it is not already available in the PATH
-        uses: julia-actions/setup-julia@v1
+        uses: julia-actions/setup-julia@v2
         with:
           version: '1'
           arch: ${{ runner.arch }}
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,45 @@
+name: CompatHelper
+on:
+  schedule:
+    # Check for dep updates on run at midnight on the first day of every month
+    - cron: 0 0 1 * *
+  workflow_dispatch:
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if Julia is already available in the PATH
+        id: julia_in_path
+        run: which julia
+        continue-on-error: true
+      - name: Install Julia, but only if it is not already available in the PATH
+        uses: julia-actions/setup-julia@v1
+        with:
+          version: '1'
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,45 @@
...
+          version = "3"
+          Pkg.add(; name, uuid, version)
+        shell: julia --color=yes {0}
+      - name: "Run CompatHelper"
+        run: |
+          import CompatHelper
+          CompatHelper.main()
+        shell: julia --color=yes {0}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tdivoll, line 44: Did you make new secrets or do we need to add the ones we already have in the repo?

.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if Julia is already available in the PATH
+        id: julia_in_path
+        run: which julia
+        continue-on-error: true
+      - name: Install Julia, but only if it is not already available in the PATH
+        uses: julia-actions/setup-julia@v1
```
tdivoll, line 19: update this to `setup-julia@v2`? Or let the dependabot flag it for updating?

tdivoll (changes requested):

.github/workflows/CompatHelper.yml
cpaniaguam: I guess we can just change it here and now.

.github/workflows/CompatHelper.yml
cpaniaguam: I'll add a new one.

cpaniaguam: No, it'll use the secret for the documentation build.

.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if Julia is already available in the PATH
+        id: julia_in_path
+        run: which julia
+        continue-on-error: true
+      - name: Install Julia, but only if it is not already available in the PATH
+        uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
```
tdivoll, line 21: Maybe this needs to be `version: '2'`? It seems redundant to declaring `v2` on line 19.

tdivoll (approved):

.github/workflows/CompatHelper.yml
cpaniaguam: That's actually the version of Julia that the action installs. See https://github.com/julia-actions/setup-julia

PR #422, opened 2024-07-24
Update Py deps management

cpaniaguam:
Update Python deps management using requirements.txt. Looked into including it into Project.toml but it doesn't work.


tdivoll (approved):
Looks good to me, I just added it to the IFT Data Process Milestone for tracking later

Issue #424, opened 2024-08-01
Add tiling by pixel size


tdivoll:

cpaniaguam:
- [ ] Add function to find optimal (square) tile side length `l` by computing the fraction of total pixels missed for the given image.
- Use image size for total pixels and user provided initial `l`.
- Choose from `l` or `l` plus or minus 1.
- [ ] Incorporate into tiling workflows such as that in #427

danielmwatkins:
@cpaniaguam can we consider this issue closed by #427 ?

PR #429, opened 2024-09-17
Add tiling utility functions for optimal tile size calculation

cpaniaguam:
See #424.

Features: 
- `get_optimal_tile_size`
- `get_tiles`  which adjusts tiles if necessary
- test suite


src/tilingutils.jl
```diff
@@ -0,0 +1,112 @@
...
+end
+
+function get_area_missed(l::Int, dims::Tuple{Int,Int}, area::Int)::Float64
+    return 1 - prod(getfit(dims, l)) * l^2 / area
+end
+
+"""
+    get_optimal_tile_size(l0::Int, dims::Tuple{Int,Int}) -> Int
+
+Calculate the optimal tile size in the range [l0-1, l0+1] for the given size `l0` and image dimensions `dims`.
```
tdivoll, line 12: size in pixels?

src/tilingutils.jl
```diff
@@ -0,0 +1,112 @@
...
+
+
+"""
+    get_tiles(array, l)
+
+Generate a collection of tiles from an array.
+
+Unlike `TileIterator`, the function adjusts the bottom and right edges of the tile matrix if they are smaller than half the tile size `l`.
+"""
+function get_tiles(array, l)
```
tdivoll, line 85: Will users ever need to provide any of the function params? If so, maybe add docstrings for 'l', 'l0', 'dims'. If it's always called internally and just finds the right number of tiles based on image size then it's probably fine. It was a little confusing on what the difference between 'l' and 'l0' are.

src/tilingutils.jl
cpaniaguam: That's right.

src/tilingutils.jl
cpaniaguam: The use case will be for the arrays to be images.  `l0` is the _initial_, user-provided tile size, which may be updated by `get_optimal_tile_size`. The tiles have been chosen to be squares except possibly for the ones at bottom and right edges that might get bumped to slightly larger rectangles. `l` is for side `length` of the squares. I'll revise the docstrings, thanks.

tdivoll (approved):
Looks good, the updated docstrings make more sense now with the difference between `l` and `l0`.

cpaniaguam:
> Looks good, the updated docstrings make more sense now with the difference between `l` and `l0`.

Thanks Tim!

PR #432, opened 2024-09-24
feat: unsharp mask

cpaniaguam:
Extract logic for unsharp mask to specialized function leveraging built in `clamp`.

### DIFF
--- a/src/normalization.jl
+++ b/src/normalization.jl
@@ -131,6 +131,21 @@ function unsharp_mask(image_gray, smoothing_param, intensity, clampmax)
 end
 
 # For old workflow in final2020.m
+"""
+    (Deprecated)
+    unsharp_mask(image_gray, smoothing_param, intensity)
+
+Apply unsharp masking on (equalized) grayscale image to enhance its sharpness.
+
+Does not perform clamping after the smoothing step. Kept for legacy tests of IceFloeTracker.jl.
+
+# Arguments
+- `image_gray`: The input grayscale image, typically already equalized.
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity`: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
+# Returns
+The sharpened grayscale image with values clipped between 0 and `clapmax`.
+"""
 function unsharp_mask(image_gray, smoothing_param, intensity)
     image_smoothed = imfilter(image_gray, Kernel.gaussian(smoothing_param))
     image_sharpened = image_gray * (1 + intensity) .- image_smoothed * intensity
--- a/src/normalization.jl
+++ b/src/normalization.jl
@@ -109,6 +109,19 @@ function imsharpen(
     return unsharp_mask(image_equalized_gray, smoothing_param, intensity)
 end
 
+"""
+    unsharp_mask(image_equalized_gray, smoothing_param, intensity)
+
+Apply unsharp masking to an equalized grayscale image to enhance its sharpness.
+
+# Arguments
+- `image_equalized_gray: The input grayscale image that has been equalized.
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
+
+# Returns
+The sharpened grayscale image with values clipped between 0.0 and 1.0.
+"""
 function unsharp_mask(image_equalized_gray, smoothing_param, intensity)
     image_smoothed = imfilter(image_equalized_gray, Kernel.gaussian(smoothing_param))
     image_sharpened = image_equalized_gray .* (1 + intensity) .+ image_smoothed .* (-intensity)
--- a/src/normalization.jl
+++ b/src/normalization.jl
@@ -122,12 +122,19 @@ Apply unsharp masking on (equalized) grayscale ([0, `clapmax`]) image to enhance
 # Returns
 The sharpened grayscale image with values clipped between 0 and `clapmax`.
 """
-function unsharp_mask(image_equalized_gray, smoothing_param, intensity, clampmax=1)
+function unsharp_mask(image_equalized_gray, smoothing_param, intensity, clampmax)
     image_smoothed = imfilter(image_equalized_gray, Kernel.gaussian(smoothing_param))
     clamp!(image_smoothed, 0.0, clampmax)
     image_sharpened = image_equalized_gray * (1 + intensity) .- image_smoothed * intensity
     clamp!(image_sharpened, 0.0, clampmax)
     return round.(Int, image_sharpened)
+end
+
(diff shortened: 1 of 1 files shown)


src/normalization.jl
```diff
@@ -106,11 +106,26 @@ function imsharpen(
...
-    return min.(max.(image_sharpened, 0.0), 1.0)
+# Arguments
+- `image_equalized_gray: The input grayscale image that has been equalized.
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
+
+# Returns
+The sharpened grayscale image with values clipped between 0.0 and 1.0.
+"""
+function unsharp_mask(image_equalized_gray, smoothing_param, intensity)
```
hollandjg, line 125: does the gray image have to be equalized for the algorithm to work? if not, I'd rename this argument to be `image_gray`. If so, can you say why?

hollandjg (commented):
One question here

src/normalization.jl
cpaniaguam: No, it doesn't. That name was chosen as a reminder that in the intended use case workflow this operation is applied to an image that has been equalized and has values either in [0, 1] or [0, 255]. @danielmwatkins what do you think?

danielmwatkins: I think that the more general name is good! Since this one is applied to a grayscale image naming the argument image_gray makes sense.

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
+- `image_equalized_gray: The input grayscale image, typically already equalized.
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
+
+# Returns
+The sharpened grayscale image with values clipped between 0 and `clapmax`.
+"""
+function unsharp_mask(image_equalized_gray, smoothing_param, intensity, clampmax)
     image_smoothed = imfilter(image_equalized_gray, Kernel.gaussian(smoothing_param))
+    clamp!(image_smoothed, 0.0, clampmax)
```
danielmwatkins, line 127: The goal is to make sure that the input image is still within the range of allowed values, right? Allowing for the image data to be either 0-1 or from 0-255. My understanding is that gaussian smoothing cannot increase the maximum value of an array or decrease the minimum value. In what case is this first call to `clamp` needed?

cpaniaguam: I ran the algorithm on test images and was getting negative values, so clipping is necessary at least for the lower bound.

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
     image_equalized_gray = Gray.(image_equalized)
 
-    image_smoothed = imfilter(image_equalized_gray, Kernel.gaussian(smoothing_param))
+    return unsharp_mask(image_equalized_gray, smoothing_param, intensity)
+end
+
+"""
+    unsharp_mask(image_gray, smoothing_param, intensity)
+
+Apply unsharp masking on (equalized) grayscale ([0, `clapmax`]) image to enhance its sharpness.
```
hollandjg, line 115:
```suggestion
Apply unsharp masking on (equalized) grayscale ([0, `clampmax`]) image to enhance its sharpness.
```

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
 
     image_equalized_gray = Gray.(image_equalized)
 
-    image_smoothed = imfilter(image_equalized_gray, Kernel.gaussian(smoothing_param))
+    return unsharp_mask(image_equalized_gray, smoothing_param, intensity)
+end
+
+"""
+    unsharp_mask(image_gray, smoothing_param, intensity)
```
hollandjg, line 113:
```suggestion
    unsharp_mask(image_gray, smoothing_param, intensity, clampmax)
```

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
+
+Apply unsharp masking on (equalized) grayscale ([0, `clapmax`]) image to enhance its sharpness.
+
+# Arguments
+- `image_gray: The input grayscale image, typically already equalized.
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
+
```
hollandjg, line 121:
```suggestion
- `clampmax`: upper limit of intensity values in the returned image.`
```

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
+    return unsharp_mask(image_equalized_gray, smoothing_param, intensity)
+end
+
+"""
+    unsharp_mask(image_gray, smoothing_param, intensity)
+
+Apply unsharp masking on (equalized) grayscale ([0, `clapmax`]) image to enhance its sharpness.
+
+# Arguments
+- `image_gray: The input grayscale image, typically already equalized.
```
hollandjg, line 118:
```suggestion
- `image_gray`: The input grayscale image, typically already equalized.
```

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
+- `smoothing_param::Int`: The pixel radius for Gaussian blurring (typically between 1 and 10).
+- `intensity: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
```
hollandjg, line 120:
```suggestion
- `intensity`: The amount of sharpening to apply. Higher values result in more pronounced sharpening.
```

src/normalization.jl
```diff
@@ -106,11 +106,35 @@ function imsharpen(
...
+    image_sharpened = image_gray * (1 + intensity) .- image_smoothed * intensity
+    clamp!(image_sharpened, 0.0, clampmax)
+    return round.(Int, image_sharpened)
+end
 
-    image_sharpened =
-        image_equalized_gray .* (1 + intensity) .+ image_smoothed .* (-intensity)
-    return min.(max.(image_sharpened, 0.0), 1.0)
+# For old workflow in final2020.m
+function unsharp_mask(image_gray, smoothing_param, intensity)
```
hollandjg, line 134: Does this function need to keep the original docstring?

hollandjg (changes requested):
A couple more comments on the documentation, which seems to be out of date. Code looks good!

src/normalization.jl
danielmwatkins: interesting, makes sense

danielmwatkins (dismissed):

src/normalization.jl
cpaniaguam: I'll add a docstring and deprecate the method. Keeping it so old tests continue to pass.

PR #439, opened 2024-09-27
chore: update dev container

hollandjg:
Use a standard `base` environment from Microsoft with "features" for Julia and Python.

Tested in VSCode on the following hosts:
- [ ] macOS ARM64
- [x] Linux in Codespace with AMD64

Still broken:
- Setting up python environments
- GDAL is missing on macOS


.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,7 @@
...
-		// Configure properties specific to VS Code.
-		"vscode": {
-			// Add the IDs of extensions you want installed when the container is created.
-			"extensions": [
-				"julialang.language-julia"
-			]
-		}
+	"image": "mcr.microsoft.com/vscode/devcontainers/base",
+	"features": {
+		"ghcr.io/julialang/devcontainer-features/julia:1": {},
```
cpaniaguam, line 4: Does this come with the Julia extension?

hollandjg: Yes, it's built in

PR #442, opened 2024-10-09
fix: pass possible_ice_threshold to find_reflectance_peaks in find_ice_labels function

cpaniaguam:


cpaniaguam:
Merging as this is a simple change.
@tdivoll

PR #447, opened 2024-10-17
fix: drop eval

cpaniaguam:


tdivoll:
Can this be closed as fixed by #456 ?

PR #451, opened 2024-10-18
Tiled workflow feats

cpaniaguam:
This adds several features needed for the tiled workflow in `master.m` from [line 593](https://github.com/WilhelmusLab/ice-floe-tracker/blob/0076a08836b2de271f254bd2a6e3cc852acaf5f1/existing_code/MASTER.m#L593) to [line 657](https://github.com/WilhelmusLab/ice-floe-tracker/blob/0076a08836b2de271f254bd2a6e3cc852acaf5f1/existing_code/MASTER.m#L657).

A similar workflow exists from [line 882](https://github.com/WilhelmusLab/ice-floe-tracker/blob/0076a08836b2de271f254bd2a6e3cc852acaf5f1/existing_code/MASTER.m#L882) to [line 944](https://github.com/WilhelmusLab/ice-floe-tracker/blob/0076a08836b2de271f254bd2a6e3cc852acaf5f1/existing_code/MASTER.m#L944).

I'd appreciate reviewers pay close attention to the implementation of the `get_nlabels` function. The logic in the original workflows is quite convoluted and this implementation seeks to simplify it. Thanks in advance!

Finally, I didn't like how `ImageConstrastAdjustment.build_histogram` or the corresponding function from `StatsBase` do their thing for this use case, so a friendlier implementation is also included.


danielmwatkins (dismissed):
The changes all look reasonable to me, nice to see the additions. It would be good to clarify the differences between find ice labels and get ice labels to see if it's possible to make them one function with arguments addressing any differences.

cpaniaguam:
> The changes all look reasonable to me, nice to see the additions. It would be good to clarify the differences between find ice labels and get ice labels to see if it's possible to make them one function with arguments addressing any differences.

@danielmwatkins `find_ice_labels` is the old implementation that's needed for the workflow in final2020.m and a lot more specific in terms of what it does; it actually does more than what's in the tiled workflow. On the other hand, `get_ice_labels` only builds the `ice_labels` mask given thresholds for each channel and it used in the tiled workflow. In fact, the functionality of `get_ice_labels` is used in find_ice_labels three times, once per relaxation attempt. I think I will refactor `find_ice_labels` so that it uses `get_ice_labels`. Would `get_ice_labels_mask` be a better name?

src/segmentation_a_direct.jl
```diff
@@ -9,22 +9,38 @@ Apply k-means segmentation to a gray image to isolate a cluster group representi
...
 )::BitMatrix
-    Random.seed!(45) # this seed generates consistent clusters for the final output
-    gray_image_height, gray_image_width = size(gray_image)
-    gray_image_1d = vec(gray_image)
-    @info("Done with reshape")
+    segmented = kmeans_segmentation(gray_image; k=k, maxiter=maxiter)
 
-    ## NOTE(tjd): this clusters into 4 classes and solves iteratively with a max of 50 iterations
+    ## Isolate ice floes and contrast from background
+    nlabel = StatsBase.mode(segmented[ice_labels])
```
danielmwatkins, line 20: Is it possible for this function to be called with an empty ice_labels vector? If so what is returned as the mode?

src/tilingutils.jl
```diff
@@ -199,8 +195,248 @@ function get_tiles(array, side_length::Int)
...
+
+"""
+    get_final(img, label, segment_mask, se_erosion, se_dilation)
+
+Final processing following the tiling workflow.
+"""
+function get_final(img, label, segment_mask, se_erosion, se_dilation)
+    img = hbreak(img)
+
+    # doesn't do much/anything
```
danielmwatkins, line 414: I'm fine with leaving the "label==1" out entirely now that we know it doesn't do anything.

src/segmentation_a_direct.jl
cpaniaguam: No. An error would be thrown:
```julia
julia> using StatsBase: mode

julia> mode([])
ERROR: ArgumentError: mode is not defined for empty collections
Stacktrace:
 [1] mode(a::Vector{Any})
   @ StatsBase ~/.julia/packages/StatsBase/ebrT3/src/scalarstats.jl:112
 [2] top-level scope
   @ REPL[4]:1

julia>
```

danielmwatkins (changes requested):
I think this is almost there. I don't think we need to include the label==1 stuff where we came to the conclusion that it doesn't affect the results and was likely a mistake in the original code.

src/segmentation_a_direct.jl
danielmwatkins: OK, that's good to know. We should take a look a few steps earlier in the flow chart to check what happens when ice_labels comes out empty. We know it happens sometimes such as when the image is fully cloudy, so I think there is already a step that addresses this.

src/tilingutils.jl
```diff
@@ -199,8 +195,248 @@ function get_tiles(array, side_length::Int)
...
+
+    marker = branch(img)
+
+    mask = MorphSE.erode(marker, se_erosion)
+    mask = MorphSE.dilate(mask, se_dilation)
+
+    # doesnt do much
+    if label == 1
+        mask[1] = false
+    end
```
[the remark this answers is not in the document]
cpaniaguam: We could. The thing is that sometimes `labels` can be `2` or `3` (as I've seen in using test images). In those cases, no updates would be done.

src/segmentation_a_direct.jl
cpaniaguam: We could wrap it so it would return some sentinel value like `missing` or `nothing` and decide on a course of action when that happens.

PR #458, opened 2024-10-23
Support labeled segmented images

hollandjg:
Add support for integer floe masks, in addition to boolean ones. 

The primary place this interfaces with the existing code is when adding the cropped floe images to the dataframe during the tracking script. The integer-labeled floes from the images are cast to `BitMatrix` and the rest of the code should work as before.


src/tracker/tracker-funcs.jl
```diff
@@ -507,13 +507,13 @@ Convert the centroid coordinates from row and column to latitude and longitude d
 """
 function convertcentroid!(propdf, latlondata, colstodrop)
     latitude, longitude = [
-        [latlondata[c][x, y] for
+        [latlondata[c][Integer(round(x)), Integer(round(y))] for
```
hollandjg, line 510: Note that this Integer cast is required for the lat-lon data which comes from EBSEG – it produces centroids with subpixel precision.
It think it would make more sense to do a proper affine transformation here, rather than getting a mapping of Xs, Ys to latitudes and longitudes. But that would be a thing for later, once we have a Julia-specific wrapper for PROJ in the project and we're not using python.

cpaniaguam: You should use `round` with some care in Julia:
```julia
julia> Int(round(6.5))
6

julia> Int(round(7.5))
8
```
If you'd like to round the "usual" way with the cutoff at 0.5, you should use `RoundNearestTiesUp` as the rounding mode. https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/ae20877a5833509afd8a356aa457059906830994/src/histogram_equalization.jl#L2)

src/tracker/tracker-funcs.jl
```diff
@@ -507,13 +507,13 @@ Convert the centroid coordinates from row and column to latitude and longitude d
...
          (x, y) in zip(propdf.row_centroid, propdf.col_centroid)] for
         c in ["latitude", "longitude"]
     ]
 
     x, y = [
-        [latlondata[c][z] for z in V] for
+        [latlondata[c][Integer(round(z))] for z in V] for
```
cpaniaguam, line 516: Using `Integer` for casting seems to work. However,  `Integer` [is an abstract type](https://docs.julialang.org/en/v1/base/numbers/#Core.Integer). For casting to concrete types, consider using `Int`.

test/test-regionprops-labeled.jl
```diff
@@ -0,0 +1,260 @@
...
+
+    ]
+
+    @info "testing props with label"
+    props_with_label = DataFrame(label=Int[], min_row=Int[], min_col=Int[], max_row=Int[], max_col=Int[])
+    push!(props_with_label, (1, 1, 1, 4, 4))
+    push!(props_with_label, (2, 1, 4, 5, 8))
+    push!(props_with_label, (3, 1, 8, 6, 13))
+    push!(props_with_label, (4, 5, 1, 10, 7))
+    push!(props_with_label, (5, 5, 1, 10, 7))
```
cpaniaguam, line 34:
```suggestion
    props_with_label = DataFrame(
        label=[1, 1, 1, 4, 4],
        min_row=[2, 1, 4, 5, 8],
        min_col=[3, 1, 8, 6, 13],
        max_row=[4, 5, 1, 10, 7],
        max_col=[5, 5, 1, 10, 7],
    )
```

test/test-regionprops-labeled.jl
```diff
@@ -0,0 +1,260 @@
...
+
+
+
+    @info "testing props with only label"
+    props_only_label = DataFrame(label=Int[])
+    push!(props_only_label, (1,))
+    push!(props_only_label, (2,))
+    push!(props_only_label, (3,))
+    push!(props_only_label, (4,))
+    push!(props_only_label, (5,))
```
cpaniaguam, line 131:
```suggestion
    props_only_label = DataFrame(label=1:5)
```

test/test-regionprops-labeled.jl
```diff
@@ -0,0 +1,260 @@
...
+        0 0 0 0 0 0 0
+    ]
+
+    @info "testing props without label"
+    props_without_label = DataFrame(min_row=Int[], min_col=Int[], max_row=Int[], max_col=Int[])
+    push!(props_without_label, (1, 1, 4, 4))
+    push!(props_without_label, (1, 4, 5, 8))
+    push!(props_without_label, (1, 8, 6, 13))
+    push!(props_without_label, (5, 1, 10, 7))
+    push!(props_without_label, (5, 1, 10, 7))
```
cpaniaguam, line 85: See suggestion above.

src/regionprops.jl
```diff
@@ -202,12 +246,67 @@ function cropfloe(floesimg::BitMatrix, props::DataFrame, i::Int64)
     return prefloe
 end
 
+"""
+    cropfloe(floesimg, min_row, min_col, max_row, max_col, label)
+
+Crops the floe from `floesimg` with the label `label`, returning the region bounded by `min_row`, `min_col`, `max_row`, `max_col`, and converting to a BitMatrix.
+"""
+function cropfloe(floesimg::Matrix{I}, min_row::J, min_col::J, max_row::J, max_col::J, label::I)  where {I<:Integer, J<:Integer}
```
cpaniaguam, line 254: Because `I` and `J` are the same type, they can be combined into one.
```suggestion
function cropfloe(floesimg::Matrix{T}, min_row::T, min_col::T, max_row::T, max_col::T, label::T)  where {T<:Integer}
```

src/tracker/tracker-funcs.jl
hollandjg: Thanks for the tip.

I think the rounding to the nearest even number in case of a tie is fine in this case – the numbers are floats, so the likelihood that we're going to be on precisely 0.5 is slim. 

If there's some situation where halves come out of the centroiding algorithm regularly, then reducing the bias by rounding to the nearest even number will probably be what we want.

test/test-regionprops-labeled.jl
hollandjg: Each of those entries is a `(label, min_row, min_col, max_row, max_col)` record, rather than being the columns, so adding them one row at a time with push is what I intended.

test/test-regionprops-labeled.jl
hollandjg: Nice!

test/test-regionprops-labeled.jl
cpaniaguam: Then you could do a sort of "transposition".

src/regionprops.jl
hollandjg: Not quite. They might be different integer types. The matrix values `I` have to be the same concrete type as the labels. But the column and row numbers might be a different concrete integer type. 

Concrete: You might have a `UInt8` for the labels, with maximum of 256 individual labels for 255 different floes and a background, but you'll probably need more than that for the row and column indices – we're likely in the `UInt16` range with up to 65563 pixels on a side.

src/tracker/tracker-funcs.jl
hollandjg: Good catch! It looks like `Integer(foo)` defaults to `Int64(foo)` on my machine. The docs [^1] say that `Base.Int` is the default type for array indices, and its concrete type depends on the system word size. Let's go for that.

[^1]: https://docs.julialang.org/en/v1/base/numbers/#:~:text=This%20is%20the%20default%20type%20of%20most%20integer%20literals%20and%20is%20an%20alias%20for%20either%20Int32%20or%20Int64%2C%20depending%20on%20Sys.WORD_SIZE.%20It%20is%20the%20type%20returned%20by%20functions%20such%20as%20length%2C%20and%20the%20standard%20type%20for%20indexing%20arrays.

src/regionprops.jl
cpaniaguam: I see your point, but I am looking at this a bit more abstractly. For the purposes of multiple dispatch and argument types in function signatures, specially when using the same abstract type, the directive `where {I<:Integer, J<:Integer}` has the same effect as `where {T<:Integer}` and use `T` for both `I` and `J`. They do the same thing but the version in the suggestion is shorter.

src/bwtraceboundary.jl
```diff
@@ -54,7 +54,7 @@ julia> boundary[3]
 ```
 """
 function bwtraceboundary(
-    image::Union{Matrix{Int64},Matrix{Float64},T};
+    image::Union{Matrix{UInt8},Matrix{Int64},Matrix{Float64},T};
     P0::Union{Tuple{Int,Int},CartesianIndex{2},Nothing}=nothing,
     closed::Bool=true,
 ) where {T<:AbstractArray{Bool,2}}
```
cpaniaguam, line 60: Because both `UInt8` and `Int64` are subtypes of `Integer`, the function signature could be simplified to something like
```julia
function foo(
    image::AbstractArray{T,2};
    P0::Union{Tuple{Integer,Integer},CartesianIndex{2},Nothing}=nothing,
    closed::Bool=true,
) where {T<:Union{Integer,AbstractFloat,Bool}}
```

src/regionprops.jl
```diff
@@ -202,12 +246,67 @@ function cropfloe(floesimg::BitMatrix, props::DataFrame, i::Int64)
...
+    #= 
+    Crop the floe using bounding box data in props.
+    Note: Using a view of the cropped floe was considered but if there were multiple components in the cropped floe, the source array with the floes would be modified. =#
+    prefloe = floesimg[min_row:max_row, min_col:max_col]
+    @debug "prefloe: $prefloe"
+
+    #= Remove any pixels not corresponding to that numbered floe 
+    (each segment has a different integer) =#
+    floe_area = prefloe .== label
+    @debug "mask: $floe_area"
```
cpaniaguam, line 264: 👍

src/regionprops.jl
```diff
@@ -202,12 +246,67 @@ function cropfloe(floesimg::BitMatrix, props::DataFrame, i::Int64)
...
+    cols = 1
+    row_sums = count(floe_area, dims=rows)
+    col_sums = count(floe_area, dims=cols)
+    @debug "row_sums: $row_sums, col_sums: $col_sums"
+
+    min_row = findfirst(nonzero, row_sums)[cols]
+    min_col = findfirst(nonzero, col_sums)[rows]
+    max_row = findlast(nonzero, row_sums)[cols]
+    max_col = findlast(nonzero, col_sums)[rows]
+    @debug "($min_row, $min_col), ($max_row, $max_col)"
```
cpaniaguam, line 292: This is interesting! Consider moving it to a function of its own (`get_bbox`?).

src/regionprops.jl
```diff
@@ -202,12 +246,67 @@ function cropfloe(floesimg::BitMatrix, props::DataFrame, i::Int64)
...
+
+    floe_area_cropped = floe_area[min_row:max_row, min_col:max_col]
```
cpaniaguam, line 294: Here a different method of `cropfloe` could be used making it a bit more expressive.

src/regionprops.jl
```diff
@@ -180,16 +180,60 @@ function getbboxcolumns(props::DataFrame)
...
+                floesimg_bitmatrix,
+                props_row.min_row,
+                props_row.min_col,
+                props_row.max_row,
+                props_row.max_col
+            )
+        end
+    elseif "label" in colnames
+        return cropfloe(floesimg, props_row.label)
+    end
```
cpaniaguam, line 224: Consider a flatter approach:
```suggestion
    colnames = Set(names(props_row)
    bbox = ["min_row", "min_col", "max_row", "max_col"]
    bbox_label = [bbox..., "label"]

    issubset(Set(bbox_label), colnames) && return cropfloe(
        floesimg,
        props_row[bbox_label]...
    )

    issubset(Set(bbox), colnames) && return cropfloe(
        floesimg .> 0,
        props_row[bbox]...
    )

    "label" in colnames && return cropfloe(floesimg, props_row.label)
```

src/regionprops.jl
hollandjg: I don't think that's correct. They don't do the same thing, because the `{T<:Integer}` means `T` is going to be a single type, which is a subtype of `Integer`, and all the arguments are going to be of that type. So it will match to `a::Int8, b::Int8, c::Int8` but not `a::int8, b::Int16, c::Int32`.

I did a quick example here: https://colab.research.google.com/drive/1K35IImi2g8AF1KXfp59eckrnjyCCrEc2?usp=sharing

src/regionprops.jl
hollandjg: I like the idea, and I've introduced the sets. 

I'd prefer to not have the list of column names doing double duty as both the set of names to check and the ordering of the columns for the output – that could lead to some really weird errors in future if someone who doesn't know that the order of the column names in the list is important and decides to (say) put them in alphabetical order. Having the order of the function arguments encoded in a list would also make it more difficult for refactoring software to help us refactor the functions if we need to in the future. So I think it's best to write those out explicitly. The performance loss will be minimal if anything, and I think this makes the code more explicit.

It would be a different story if we could use keyword arguments here, but that's not possible for the multiple-dispatch kind of function we're using.

hollandjg: I like the idea, and I've introduced the sets. 

I'd prefer to not have the list of column names doing double duty as both the set of names to check and the ordering of the columns for the output – that could lead to some really weird errors in future if someone who doesn't know that the order of the column names in the list is important and decides to (say) put them in alphabetical order. Having the order of the function arguments encoded in a list would also make it more difficult for refactoring software to help us refactor the functions if we need to in the future. So I think it's best to write those out explicitly. The performance loss will be minimal if anything, and I think this makes the code more explicit.

It would be a different story if we could use keyword arguments here, but that's not possible for the multiple-dispatch kind of function we're using.

src/regionprops.jl
hollandjg: How do you mean?

test/test-regionprops-labeled.jl
hollandjg: How about this?

src/regionprops.jl
hollandjg: It would pretty much be the whole function, from line 286 through 297, and I'm not sure we use it anywhere else, so I don't know that abstracting it would make the code more readable. Is there somewhere else we might call it? The old way of cropping the floes (with `bwareamaxfilt`) is subtly different, so we can't reuse it there.

src/bwtraceboundary.jl
hollandjg: I like this suggestion, but I think it might be best in a separate PR. I'll make an issue. #460

test/test-regionprops-labeled.jl
cpaniaguam: I think you might have forgotten to paste in something in there.

cpaniaguam: I was thinking this
```julia
props_with_label = DataFrame(;
    label=1:5,
    min_row=[1, 1, 1, 5, 5],
    min_col=[1, 4, 8, 1, 1],
    max_row=[4, 5, 6, 10, 10],
    max_col=[4, 8, 13, 7, 7],
)
```

avoiding the extra call to push and passing the data directly to the constructor.

hollandjg: I don't think that's a good idea. Each testcase is a single row, and my intention was that you can easily add a new testcase in any position in the dataframe, just by adding a new row. 

By splitting it up by column: 

- there is one fewer function call, yes, but
- it makes it much harder to add a new testcase, because you have to modify multiple lines, touching columns which don't line up, and 
- it makes debugging much more difficult, because understanding what the code is doing for each case requires parsing multiple lines, working out which value in which column is the relevant one, matching that to the range in the first column.

I don't think that future developers will appreciate the added complexity. I would think that the single additional function call makes no meaningful difference to the speed of running these testcases.

If there's a way to pass the rows directly into the constructor then I'm happy to change the code to do that, but when I looked there wasn't.

cpaniaguam: Very good points, thanks!

In that use case one can still pass the rows to the constructor. Consider this:
```julia
cases = [
    1 1 1 4 4
    2 1 4 5 8
    3 1 8 6 13
    4 5 1 10 7
    5 5 1 10 7
]

props_with_label = DataFrame(cases, [:label, :min_row, :min_col, :max_row, :max_col])
```

hollandjg: ✅

test/test-regionprops-labeled.jl
hollandjg: ✅

src/regionprops.jl
cpaniaguam: This has been on my mind for the last couple of days. I recalled that individual floe masks are needed for the tracker. One is able to get all of them in one go with the [`bbox` prop from `regionprops`](https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.regionprops:~:text=as%20the%20region.-,bboxtuple,-Bounding%20box%20(min_row ) as I did [here](https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/939866c522f0e27c86185a11a2017634b5e6ac75/src/regionprops.jl#L87). Did you consider that?

hollandjg: I think if we do that, we should avoid adding a new python dependency, and instead use something like https://github.com/zygmuntszpak/ImageComponentAnalysis.jl/blob/0d4503f5a4b2bed647f99a40adc485d5413c8afb/src/algorithms/bounding_box.jl#L1-L26

But do we want to do that now? It looks to be doing the same thing, but it's assuming it's going to loop over a bunch of labels, rather than providing us the ability to do just one. Could we put this in as an issue and come back to it later?

cpaniaguam: The way that I see this is that one would be processing one now but later one would like to aggregate them all for the tracker step. Getting the bbox from regionprops avoids that extra overhead entirely.

cpaniaguam: @mirestrepo It's from an existing dependency (`skimage`). We've had conversations around how to eliminate it but it's not easy as there are props we need for the tracker that we can't get from the Julia ecosystem without extra work.

hollandjg: After some checks, I've removed this bit of code, as we won't need it.

PR #464, opened 2024-10-25
Replace PyCall with PythonCall

hollandjg:
When working on the pipeline, I was having trouble with building the docker because of the PyCall dependency. I wanted to see how easy it was to replace PyCall with PythonCall which has a better way of handling dependencies.

Turns out it's also hard:
- PythonCall doesn't have an obvious way to import a custom file (`latlon.py`)
- We have to package that separately, and the import it using Pip in Conda.
- Using Pip in Conda reinstalls the dependencies from the package installed using Pip.
- We'd need to make latlon.py into a conda package and push it to a registry to make this easier. 

I don't think we should merge this PR – it just makes things complicated in a different way. I think we should do our best to remove the python dependencies from this package. 

- LatLon: 
  - We could replace it with something based on Proj.jl (preferred)
  - We could have a separate command-line tool which does the reprojection and read the values in from a flat file
- Scikit-image: region props:
  - We could have a small command line tool based on a part of ebseg which takes an integer labeled image and produces the region props as we need them
  - We could have a small command line tool (new) which takes in a binary mask and labels the contiguous blocks using integers, and then feed that into the first command line tool.
  - Then we read all the region props back into the julia pipeline
- Scikit-image: Exposure correction:
  - we could have a small command line tool (new) which does the exposure correction on a full image, and then pass that image back into the main pipeline.

I tried to get around this by updating the Conda environments, but found that by default the PyCall installs python3.12 which isn't compatible with our dependencies. https://github.com/WilhelmusLab/IceFloeTracker.jl/issues/465

Not sure how to proceed here.


tdivoll:
Possibly Conda.jl is another alternative, but again resolving python dependencies with newer versions of python is always a challenge.

hollandjg:
We're not going to merge this.
The notes should go into an issue.
And we'll try to reduce the python dependencies step-by-step.

hollandjg:
Closing this now. The notes are in #469.

PR #467, opened 2024-10-28
Fixing error in latlon.py

danielmwatkins:
I found a spot in latlon.py where I mixed up the row-col ordering and x-y ordering between Numpy and Rasterio.


danielmwatkins:
I've added an image and two CSV files for setting up a test for latlon.py. When you run get_latlon, you get a dictionary with keys 'latitude', 'longitude', 'X', and 'Y'. I added two CSV files, latitude.csv and longitude.csv, so that the columns of each are X and the indices are Y, and the value at position x, y is the latitude or longitude for that index pair.

src/latlon.py
```diff
@@ -12,14 +12,18 @@ def getlatlon(imgpath):
     im = rasterio.open(imgpath)
     crs = im.crs.__str__()  # Coordinate reference system code
     nrows, ncols = im.shape
-    rows, cols = np.meshgrid(np.arange(nrows), np.arange(ncols))
+    cols, rows = np.meshgrid(np.arange(ncols), np.arange(nrows))
     xs, ys = rasterio.transform.xy(im.transform, rows, cols)
+    xs = np.array(xs)
+    ys = np.array(ys)
```
cpaniaguam, line 18: Is the original data type in `xs`, and `ys` different than `np.array`? Again, just curious.

src/latlon.py
```diff
@@ -12,14 +12,18 @@ def getlatlon(imgpath):
     im = rasterio.open(imgpath)
     crs = im.crs.__str__()  # Coordinate reference system code
     nrows, ncols = im.shape
-    rows, cols = np.meshgrid(np.arange(nrows), np.arange(ncols))
+    cols, rows = np.meshgrid(np.arange(ncols), np.arange(nrows))
```
cpaniaguam, line 15: I guess this is doing the same as before? Why write this way?  Just curious.

src/latlon.py
```diff
@@ -12,14 +12,18 @@ def getlatlon(imgpath):
...
     xs, ys = rasterio.transform.xy(im.transform, rows, cols)
+    xs = np.array(xs)
+    ys = np.array(ys)
+    
     # X and Y are the 1D index vectors
-    X = np.array(xs)[:, 0]
-    Y = np.array(ys)[0, :]
+    X = np.array(xs)[0, :]
+    Y = np.array(ys)[:, 0]
```
cpaniaguam, line 22: Because of the earlier casting, those operations could be dropped here or the earlier ones removed.

src/latlon.py
```diff
@@ -12,14 +12,18 @@ def getlatlon(imgpath):
...
     ps_to_ll = Transformer.from_crs(im.crs, "WGS84", always_xy=True)
     lons, lats = ps_to_ll.transform(np.ravel(xs), np.ravel(ys))
+    
     # longitude and latitude are 2D index arrays
     longitude = np.reshape(lons, (nrows, ncols))
     latitude = np.reshape(lats, (nrows, ncols))
-    return {"crs": crs, "longitude": longitude, "latitude": latitude, "X": X, "Y": Y}
+    return {"crs": crs, "longitude": longitude, "latitude": latitude, "X": X, "Y": Y}
```
cpaniaguam, line 29: This line is the same too but GitHub is picky about not leaving an empty line and the end of the file.

cpaniaguam (changes requested):
A couple of questions and one suggested cleanup. Are you adding a file with the actual test exercising the function?

src/latlon.py
danielmwatkins: Yes, it's a list

src/latlon.py
danielmwatkins: It's not the same as before, that's the point -- in the case where nrows == ncols, yes, it's just a change in names, but in general nrows != ncols. The output of mesh grid is two matrices that mark the indices, and for a square matrix the first matrix is the transpose of the other.

src/latlon.py
danielmwatkins: Done, this line was supposed to get changed due to the cast earlier.

src/latlon.py
danielmwatkins: Regarding the test -- no, I did not write a test function. The way the Julia test stuff works is still pretty opaque to me. I just set up the test data so that it would be more straightforward to write the test function later.

src/latlon.py
cpaniaguam: I'll add the test. Thanks for the data!

src/latlon.py
danielmwatkins: Funny how it's picky about that here, but when I delete the empty line and save the file, it's not enough to register it as a change when I check "git status"

cpaniaguam (approved):
This is looking good to me now. @danielmwatkins Feel free to take a look at the test file and associated data.

cpaniaguam (approved):
@danielmwatkins All good now!

PR #480, opened 2024-11-06
479 imbrighten

### LINKED ISSUE #479: imbrighten
- [x] get_brighten_mask
- [x] imbrighten
- [x] tests

cpaniaguam:
Closes #479.

### DIFF
--- a/src/IceFloeTracker.jl
+++ b/src/IceFloeTracker.jl
@@ -71,6 +71,7 @@ include("branch.jl")
 include("special_strels.jl")
 include("tilingutils.jl")
 include("histogram_equalization.jl")
+include("brighten.jl")
 
 
 const sk_measure = PyNULL()
--- a/src/brighten.jl
+++ b/src/brighten.jl
@@ -1,0 +1,28 @@
+"""
+    get_brighten_mask(equalized_gray_reconstructed_img, gamma_green)
+# Arguments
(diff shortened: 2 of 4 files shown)


danielmwatkins (approved):

src/IceFloeTracker.jl
```diff
@@ -71,7 +71,7 @@ include("branch.jl")
 include("special_strels.jl")
 include("tilingutils.jl")
 include("histogram_equalization.jl")
-include("imadjust.jl")
```
tdivoll, line 74: I think the CI fails because of this deletion. Maybe keep the `include("imadjust.jl")` and add include("brighten.jl")? It looks like there is still an `imadjust.jl` script, suggesting that maybe we need both?

tdivoll (changes requested):

src/IceFloeTracker.jl
cpaniaguam: Must have done that trying to resolve merge conflicts. I think it's fixed now. Thanks.

cpaniaguam:
It's missing the call to the @test macro. I'll fix that!
https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/d0926da71d349aeef2cfaf6fde31bb94b774b9f3/test/test-imadjust.jl#L4

tdivoll (approved):
Tests passing now, looks good!

PR #483, opened 2024-11-06
feat: implement parallel processing for gradient magnitude computation

cpaniaguam:


src/segmentation_a_direct.jl
```diff
@@ -112,7 +112,9 @@ Compute the gradient magnitude of an image using the Sobel operator.
 """
 function imgradientmag(img)
     h = centered([-1 0 1; -2 0 2; -1 0 1]')
```
tdivoll, line 114: Is this a Sobel kernel? If so maybe just add a comment after to indicate so

src/segmentation_a_direct.jl
```diff
@@ -112,7 +112,9 @@ Compute the gradient magnitude of an image using the Sobel operator.
...
-    Gx = imfilter(img, h', "replicate")
-    Gy = imfilter(img, h, "replicate")
+    Gx_future = Threads.@spawn IceFloeTracker.imfilter(img, h', "replicate")
```
tdivoll, line 115: Why does Gx_future appluy to `h'`, but Gy_future applies to `h`? Just wanted to double check that it's not a typo.

tdivoll (changes requested):
Just a question to make sure it's not a typo, and suggest adding a comment about the filter applied in `h`.

src/segmentation_a_direct.jl
cpaniaguam: Yes, I noted that in the docstring just a couple of lines earlier.

src/segmentation_a_direct.jl
cpaniaguam: That is how the algo works in MATLAB -- see the code in the `imgradientxy.m` function.

danielmwatkins (approved):

PR #487, opened 2024-11-07
feat: add fillholes! get_segment_mask

cpaniaguam:


tdivoll (approved):
This looks like some good little abstractions and one-purpose functions.

PR #489, opened 2024-11-07
Impose minima and related functions

cpaniaguam:


danielmwatkins (dismissed):
I didn't see any docstrings, so do add those if these are going to be functions that a user can access

src/utils.jl
```diff
@@ -120,6 +120,38 @@ function imextendedmin(img::AbstractArray; h::Int=2, conn::Int=2)::BitMatrix
...
 end
 
+"""
+    impose_minima(I::AbstractArray{T}, BW::AbstractArray{Bool}) where {T<:Integer}
+
+Use morphological reconstruction to enforce minima on the input image `I` at the positions where the binary mask `BW` is non-zero.
+
+It supports both integer and grayscale images using different implementations for each.
+"""
+function impose_minima(I::AbstractArray{T}, BW::AbstractArray{Bool}) where {T<:Integer}
```
tdivoll, line 130: minor detail, but consider adding a comment after `where {T<:Integer} #color image`. 
Then below on line 139, consider `where {T<:AbstractFloat} #gray image`. 
It's clear in the docstring, so maybe not needed.

src/utils.jl
```diff
@@ -120,6 +120,38 @@ function imextendedmin(img::AbstractArray; h::Int=2, conn::Int=2)::BitMatrix
...
+    marker = 255 .* BW
+    mask = imcomplement(min.(I .+ 1, 255 .- marker))
+    reconstructed = sk_morphology.reconstruction(marker, mask)
```
tdivoll, line 133: Is this `reconstruction` from skimage closer to the matlab implementation compared to the reconstruction we were using before?

tdivoll (approved):
Just suggestions in this one, otherwise looks good!

PR #502, opened 2024-11-07
501 get ice labels with relaxation

cpaniaguam:


src/find_ice_labels.jl
```diff
@@ -84,3 +84,172 @@ function find_ice_labels(
...
+    relaxed_thresholds = [band_7_threshold_relaxed, pksb.locs[2], pksc.locs[2]]
+    ice_labels = get_ice_labels_mask(ref_img, relaxed_thresholds, factor)
+
+    sum(ice_labels) > 0 && return StatsBase.mode(morph_residue_labels[ice_labels])
+
+    # Final relaxation
+    mask_b = b .> band_2_threshold
+    sum(mask_b) > 0 && return StatsBase.mode(morph_residue_labels[mask_b])
+
+    # TODO: Should a fallback value be added? Return nothing if no ice is found? return 1? throw error?
```
danielmwatkins, line 173: What kind of missing value flags have been used elsewhere? Would it work to use a NaN, and later only run the segmentation if the icelabels is not nan?

cpaniaguam: @danielmwatkins In Julia people use [`missing`](https://docs.julialang.org/en/v1/manual/missing/#missing) for such use cases. It essentially behaves like NaN and one can test for "missingness" with `ismissing`
```julia
julia> mymode = missing
missing

julia> ismissing(mymode)
true
```
In our use case, if no labels are found we could return `missing` as a fallback and handle this afterwards.

src/find_ice_labels.jl
```diff
@@ -84,3 +84,172 @@ function find_ice_labels(
...
+end
+
+# TODO: Add tests
```
tdivoll, line 176: delete line 176, it seems like `test/test-tiled-ice-labels.jl` covers this script?

src/find_ice_labels.jl
```diff
@@ -84,3 +84,172 @@ function find_ice_labels(
...
+    cv = [float64.(cv[i, :, :]) .* factor for i in 1:3]
+    mask_ice_band_7 = cv[1] .< thresholds[1]
+    mask_ice_band_2 = cv[2] .> thresholds[2]
+    mask_ice_band_1 = cv[3] .> thresholds[3]
+    mask = mask_ice_band_7 .* mask_ice_band_2 .* mask_ice_band_1
+    @debug "Found $(sum(mask)) ice pixels"
+    return mask
+end
+
+# TODO: add test
```
tdivoll, line 112: delete line 112, it seems like `test/test-tiled-ice-labels.jl` covers this script?

src/find_ice_labels.jl
```diff
@@ -84,3 +84,172 @@ function find_ice_labels(
...
+
+    # TODO: make this conditional on input args
+    order = sortperm(heights; rev=true)
+    locs, heights = locs[order], heights[order]
+
+    return (locs=locs, heights=heights)
+end
+
+
+function get_ice_labels_mask(ref_img::Matrix{RGB{N0f8}}, thresholds, factor=1)
```
tdivoll, line 101: Will using N0f8 here cause any problems?

tdivoll (changes requested):
Great abstractions, this finding peaks and ice labels was a bear! Just some questions for clarification.

src/find_ice_labels.jl
danielmwatkins: I think that's a good idea -- if there is no ice in the image, then it should skip the segmentation section since there's nothing to detect.

danielmwatkins (dismissed):

danielmwatkins (approved):
Looks good to me!

PR #504, opened 2024-11-09
feat: preprocess with tiling

cpaniaguam:
Orchestration function for preprocessing with tiling.

- [x] Add tests


danielmwatkins (dismissed):
Outlining the steps is helpful. Is there a way to easily turn the @info logging on and off if the user were to be running a batch of images?

danielmwatkins (approved):

danielmwatkins (approved):

src/histogram_equalization.jl
```diff
@@ -127,6 +132,18 @@ function get_rgb_channels(img)
...
 
+"""
+    rgb2gray(rgbchannels::Array{Float64, 3})
+
+Convert an array of RGB channel data to grayscale in the range [0, 255].
+"""
+function rgb2gray(rgbchannels::Array{Float64,3})
+    r, g, b = [to_uint8(rgbchannels[:, :, i]) for i in 1:3]
+    # Reusing the r array to store the equalized gray image
+    r .= to_uint8(0.2989 * r .+ 0.5870 * g .+ 0.1140 * b)
```
tdivoll, line 143: Where do these specific numbers come from? 0.2989, 0.5870, 0.1140

src/histogram_equalization.jl
```diff
@@ -287,7 +304,7 @@ end
 Histogram equalization of `img` using `nbins` bins.
 """
 function histeq(img::S; nbins=64)::S where {S<:AbstractArray{<:Integer}}
```
tdivoll, line 306: Does this always need to be set to 64 bins?

src/ice_masks.jl
```diff
@@ -56,15 +144,13 @@ function get_ice_masks(
...
+    fc_landmasked = apply_landmask(falsecolor_image, .!landmask)
 
-    Threads.@threads for tile in tiles
-        #  Conditionally update binarized_tiling as its not used in some workflows
-        if binarize
-            binarized_tiling[tile...] .= imbinarize(morph_residue[tile...])
-        end
-
-        morph_residue_seglabels = kmeans_segmentation(Gray.(morph_residue[tile...] / 255))
+    # Threads.@threads
```
tdivoll, line 149: Should this be uncommented to utilize the threads?

src/ice_masks.jl
```diff
@@ -29,12 +118,10 @@ Get the ice masks from the falsecolor image and morphological residue given a pa
 - `band_1_threshold_relaxed=190`: The relaxed threshold for band 1.
 - `possible_ice_threshold=75`: The threshold for possible ice.
 - `factor=255`: normalization factor to convert images to uint8.
```
tdivoll, line 120: I think this docstring just needs a description for `k`!

tdivoll, line 120:
```suggestion
- `factor=255`: normalization factor to convert images to uint8.
- `k=3`: The number of clusters to use for k-means segmentation.
```

src/preprocess_tiling.jl
```diff
@@ -0,0 +1,210 @@
...
+
+    begin
+        @debug "Step 4: Remove clouds from equalized_gray"
+        masks = [f.(ref_img_cloudmasked) .== 0 for f in [red, green, blue]]
+        combo_mask = reduce((a, b) -> a .& b, masks)
+        equalized_gray[combo_mask] .= 0
+    end
+
+    begin
+        @debug "Step 5: unsharp_mask on equalized_gray and reconstruct"
```
tdivoll, line 121: I like these `@debug` statements to keep track of all the operations here

src/watershed.jl
```diff
@@ -16,25 +16,26 @@ function _reconst_watershed(morph_residue::Matrix{<:Integer}, se::Matrix{Bool}=s
...
-    task1 = Threads.@spawn begin
-        mr_reconst = _reconst_watershed(morph_residue)
-        mr_reconst = ImageMorphology.local_maxima(mr_reconst; connectivity=2) .> 0
-    end
+    # task1 = Threads.@spawn begin
+    mr_reconst = _reconst_watershed(morph_residue)
+    mr_reconst = ImageMorphology.local_maxima(mr_reconst; connectivity=2) .> 0
+    # end
 
     # Task 2: Calculate gradient magnitude
```
tdivoll, line 26: Maybe make an issue in the repo for threading watershed and tiling. Maybe after a hiatus, we can think about overall strategy for threading this two big operations?

tdivoll (changes requested):
Just some questions for clarifications and adding `k` to the docstrings.

src/histogram_equalization.jl
cpaniaguam: That's how the MATLAB's `rgb2gray` converts a RGB image to grayscale (uint8). See [here](https://tinyurl.com/mvceu5h8).

src/histogram_equalization.jl
cpaniaguam: Not really but that's how MATLAB's `histeq` behaves. See [here](https://tinyurl.com/mskyx9mm).

src/ice_masks.jl
cpaniaguam: Sometimes it would crash, not sure why. Maybe I was running out of memory. I just decided to leave it commented for now to ensure it'd always work.

src/watershed.jl
cpaniaguam: As in the threading case above, sometimes it would crash, but I'll add an issue noting where parallelism could be leveraged.

cpaniaguam: See #523.

src/ice_masks.jl
cpaniaguam: #524

src/histogram_equalization.jl
danielmwatkins: Let's add a comment to say the source - "Converts RGB to grayscale using a linear combination of channel brightness values. Identical to MATLAB "rgb2gray" (source link)."

cpaniaguam: I added that to the docstring.

src/histogram_equalization.jl
```diff
@@ -127,6 +132,20 @@ function get_rgb_channels(img)
...
+    rgb2gray(rgbchannels::Array{Float64, 3})
+
+Convert an array of RGB channel data to grayscale in the range [0, 255].
+
+Identical to MATLAB `rgb2gray` (https://www.mathworks.com/help/matlab/ref/rgb2gray.html).
+"""
+function rgb2gray(rgbchannels::Array{Float64,3})
+    r, g, b = [to_uint8(rgbchannels[:, :, i]) for i in 1:3]
+    # Reusing the r array to store the equalized gray image
+    r .= to_uint8(0.2989 * r .+ 0.5870 * g .+ 0.1140 * b)
```
hollandjg, line 145: Why reuse the same array? I assume you're trying to save some memory with this. Does it actually save any memory? Does it create a new array in memory anyway and then assign that to the `r` variable because `r` appears in the right hand side?

hollandjg (approved):
This looks good, just one comment on the code.

One question: does anything need to change in the CLI to make this work, or will it just slot in?

src/histogram_equalization.jl
cpaniaguam: This approach saves memory in general by preventing another allocation. In this case the RHS array is generated in memory, as you point out, and it's immediately returned so no savings. But it would have yielded savings if RHS would have been used for another operation. Thus, this or `return to_uint8(...)` does the same thing.

cpaniaguam:
> One question: does anything need to change in the CLI to make this work, or will it just slot in?

@hollandjg 
This function processes one unit of input data in the same way [`preprocess`](https://github.com/WilhelmusLab/ice-floe-tracker-pipeline/blob/113a012cfa0e4ea2c6c4b01f3b728c3df9a8529f/src/preprocess.jl#L139) does. It will need a wrapper that leverages it (like the namesake [`preprocess`](https://github.com/WilhelmusLab/ice-floe-tracker-pipeline/blob/113a012cfa0e4ea2c6c4b01f3b728c3df9a8529f/src/preprocess.jl#L217) does.

PR #509, opened 2024-11-12
fix: make sort! declaration extend Base.sort!

hollandjg:
... rather than shadowing it, which breaks the find_reflectance_peaks function in find_ice_labels.jl under some circumstances


cpaniaguam (approved):
I think I had a PR for this elsewhere but let's just get it fixed in this one.

PR #521, opened 2024-11-19
Setting up documentation pages

danielmwatkins:
I've added a placeholder logo and 3 pages where the main families of functions in the algorithm can be described.


docs/src/assets/logo.png
cpaniaguam, line 1: Cool logo!

docs/src/preprocessing.md
```diff
@@ -0,0 +1,24 @@
...
+
+```@raw html
+<img align="right" src="assets/landmask_example.png" width="300">
+```
+
+At the top, we have the original landmask TIFF, which has black and gray values. The middle image is the binary image, with land set to 0. At the bottom, we can see the dilated image using the default value of the structuring element. The default has radius 50, which results in a coastal mask of 12.5 km based on the 250 m pixel size of default MODIS imagery.
+
+## Cloud masks
+Setting thresholds for cloud mask
+
```
cpaniaguam, line 23: I guess this is a placeholder. Consider adding a `TODO: <my task>` which is easy to search for later.

src/landmask.jl
```diff
@@ -69,14 +69,16 @@ end
...
+a list of indices of pixels that are likely containing sea ice.
 
 # Arguments
 - `landmask`: bitmatrix landmask for region of interest
 - `ice_mask`: bitmatrix with ones equal to ice, zeros otherwise
 
 """
 ## NOTE(tjd): This function is called in `find_ice_labels.jl`
+## NOTE(dmw): For consistency, it would make sense to reverse the order of inputs to match landmask
 function remove_landmask(landmask::BitMatrix, ice_mask::BitMatrix)::Array{Int64}
```
cpaniaguam, line 82:
```
## NOTE(dmw): For consistency, it would make sense to reverse the order of inputs to match landmask
function remove_landmask(landmask::BitMatrix, ice_mask::BitMatrix)::Array{Int64}
```

Consider adding an issue.

cpaniaguam (approved):
Looks good. Left two suggestions.

docs/src/preprocessing.md
danielmwatkins: Yes, this is - there's an item for issue #522 for this

src/landmask.jl
danielmwatkins: That's a good idea!

PR #529, opened 2024-12-07
Feat: long tracker

cpaniaguam:
This pull request includes several changes to the `IceFloeTracker` project, focusing on improving the tracking of ice floes over multiple days, adding new functionality, and updating the corresponding tests. The most important changes include the addition of the `long_tracker` function, updates to the notebook, and the introduction of new helper functions.

New functionality:

* Added `long_tracker` function to track ice floes over multiple days, including the necessary helper functions such as `find_floe_matches`, `get_unmatched`, `get_trajectory_heads`, and others. (`src/tracker/long_tracker.jl`, `src/tracker/tracker-funcs.jl`) 
* Included the new `long_tracker` file in the main module. (`src/IceFloeTracker.jl`)

Notebook updates:

* Updated the `track-floes.ipynb` notebook to use the new `long_tracker` function and added new data processing steps and visualizations. (`notebooks/track-floes/track-floes.ipynb`) 

Testing improvements:

* Added new tests for the `long_tracker` function, including cases with gaps in data. (`test/test-long-tracker.jl`)
* Updated test setup to use the new `@ntestset` macro for better test organization. (`test/runtests.jl`, `test/test-matchcorr.jl`)

Notes about `long_tracker` 

1. This function can replace the call to `_pairfloes` within `pairfloes` which is used the pipeline workflow. 

https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/e7be057a96dc22262c6520d3243aa6fe312d7050/src/tracker/tracker.jl#L226-L246

2. It has a simpler signature. `_pairfloes` as currently defined does too much work. Besides the matching, it also adds unique ids to floes, pass times, floe masks, and psi-s curves. `long_tracker` expects all those data to be included in the input `props` object. See sample workflows in the test suite and the illustrative notebook.

TODO
- [ ] Add tie-breaking criteria for potential matches.


src/tracker/long_tracker.jl
```diff
@@ -0,0 +1,161 @@
...
+                            (ratios..., area_mismatch, corr),
+                            s,
+                            dist,
+                        )
+                        @debug "Matching floes" matching_floes
+                    end
+                end
+            end # of s for loop
+
+            # 2. Find the best match for floe r
```
hollandjg, line 136: Do you need to add a check in here that you don't have multiple matches?

src/tracker/tracker-funcs.jl
```diff
@@ -26,6 +26,7 @@ end
 Return an object of type `MatchedPairs` with an empty dataframe with the same column names as `df`, an empty dataframe with column names `area`, `majoraxis`, `minoraxis`, `convex_area`, `area_mismatch`, and `corr` for similarity ratios, and an empty vector for distances.
 """
 function MatchedPairs(df)
+    # TODO: revisit this
```
hollandjg, line 29: What's needed here?

src/tracker/long_tracker.jl
```diff
@@ -0,0 +1,161 @@
...
+        end # of for r = 1:nrow(props1)
+
+        # exit while loop if there are no more floes to match
+        @debug "Matched pairs" matched_pairs
+        isempty(matched_pairs) && break
+
+        #= Resolve collisions:
+        Are there floes in day k+1 paired with more than one
+        floe in day k? If so, keep the best matching pair and remove all others. =#
+        resolvecollisions!(matched_pairs)
```
hollandjg, line 156: I'm not sure about the name `collisions`. I don't think you're talking about physical collisions between floes, I think you're talking about multiple matches. Is that correct?

src/tracker/long_tracker.jl
```diff
@@ -0,0 +1,161 @@
...
+    long_tracker(props, condition_thresholds, mc_thresholds)
+
+Track ice floes over multiple days.
+
+Trajectories are built in two steps:
+0. Get pairs of floes in day 1 and day 2. Any unmatched floes, in both day 1 and day 2, become the "heads" of their respective trajectories.
+1. For each subsequent day, find pairs of floes for the current trajectory heads. Again, any unmatched floe in the new prop table starts a new trajectory.
+
+# Arguments
+- `props::Vector{DataFrame}`: A vector of DataFrames, each containing ice floe properties for a single day. Each DataFrame must have the following columns in the order specified:
```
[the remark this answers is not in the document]
cpaniaguam: At a point that order was needed but I removed the dependency of the order of the columns. They can now be in any order.

src/tracker/long_tracker.jl
cpaniaguam: Yes, by `collision` I am referring to two or more things being associated to the same object, not things physically clashing with each other. I believe `collision` is a standard term that captures this situation in computer science lingo. See for example https://en.wikipedia.org/wiki/Hash_collision.

src/tracker/long_tracker.jl
cpaniaguam: This is how the algorithm is implemented in the original MATLAB script. During the first phase of pairing there could be multiple matches. `resolvecollisions!` is supposed to handle that situation.

src/tracker/tracker-funcs.jl
cpaniaguam: I failed to delete that line. Thanks!

src/tracker/tracker-funcs.jl
```diff
@@ -498,6 +501,134 @@ function addfloemasks!(props::Vector{DataFrame}, imgs::Vector{<:FloeLabelsImage}
...
+"""
+function get_trajectory_heads(pairs::T) where {T<:AbstractDataFrame}
+    gdf = groupby(pairs, :uuid)
+    return combine(gdf, last)[:, names(pairs)]
+end
+
+"""
+    _swap_last_values!(df)
+
+Swap the last two values of the `area_mismatch` and `corr` columns for each group in `df`. For bookkeeping purposes for goodness of fit data during the tracking process.
```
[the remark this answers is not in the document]
cpaniaguam: Say that at step `k` you have the following trajectories:
```julia
julia> trajectories[:, [:ID, :passtime, :area_mismatch, :corr]]
6×4 DataFrame
 Row │ ID     passtime             area_mismatch  corr           
     │ Int64  DateTime             Float64?       Float64?       
─────┼───────────────────────────────────────────────────────────
   1 │     1  2022-09-14T12:00:00            0.0        0.991
   2 │     1  2022-09-15T12:00:00      missing    missing        
   3 │     2  2022-09-14T12:00:00            0.0        0.988
   4 │     2  2022-09-15T12:00:00      missing    missing        

```
Here the values of `:area_mismatch`  and `:corr` in row `i` indicate the _goodness_ of the matching between the floe data in that row and those in the `i+1`th row of that trajectory.

Assume further that at step `k+1` the algorithm suggests those trajectories should be extended with the following matches:
```julia
julia> toextend[:, [:ID, :passtime, :area_mismatch, :corr]]
6×4 DataFrame
 Row │ ID     passtime             area_mismatch  corr           
     │ Int64  DateTime             Float64?       Float64?       
─────┼───────────────────────────────────────────────────────────
   1 │     1  2022-09-16T12:00:00            0.0        0.995
   2 │     2  2022-09-16T13:00:00            0.0        0.988
    
```

How to include `toextend` within `trajectories` while preserving the meaning of `:area_mismatch`  and `:corr`? The simplest way I could think of was to just include the new rows in `toextend` within `trajectories` and sort by `ID` and then by `passtime`. But this would mess up the goodness columns.
```julia
julia> trajectories[1:8, [:ID, :passtime, :area_mismatch, :corr]]
6×4 DataFrame
 Row │ ID     passtime             area_mismatch  corr           
     │ Int64  DateTime             Float64?       Float64?       
─────┼───────────────────────────────────────────────────────────
   1 │     1  2022-09-14T12:00:00            0.0        0.991
   2 │     1  2022-09-15T12:00:00      missing    missing      
   3 │     1  2022-09-16T12:00:00            0.0        0.995  
   4 │     2  2022-09-14T12:00:00            0.0        0.988
   5 │     2  2022-09-15T12:00:00      missing    missing
   6 │     2  2022-09-16T13:00:00            0.0        0.988           

```
A call to `_swap_last_values!` fixes that distortion.

test/test-long-tracker.jl
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    @testset "Case 2" begin
+        # Unmatched floe in day 1, unmatched floe in day 2, and matches for every floe starting in day 3
+        props_test_case2 = deepcopy(_props)
+        delete!(props_test_case2[1], 1)
+        delete!(props_test_case2[2], 5)
+        trajectories = IceFloeTracker.long_tracker(props_test_case2, condition_thresholds, mc_thresholds)
+
+        # Expected: 5 trajectories, 3 of which have length 3 and 2 of which have length 2
```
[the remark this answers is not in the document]
cpaniaguam: The trajectories are numbered `1:n`. In line 73 those ID numbers and their lengths are captured.
```julia
julia> ids, counts = _imhist(IDs, unique(IDs))
([1, 2, 3, 4, 5], [3, 3, 3, 2, 2])
```
The number of trajectories can be obtained with `maximum(ids)` (see test in line 74). For the trajectory lengths, the same workflow from above is done on `counts`. Note that there are 3 `3`s and 2 `2`'s and that is checked in line 77.

src/tracker/long_tracker.jl
hollandjg: I'm aware of the meaning of the term in computing. But this is a program about physical objects which can collide too, and their locations in space are the whole point of the code, so the term "collide" may be interpreted in two different ways, and the first one which came to my mind was the physical one.

src/tracker/tracker-funcs.jl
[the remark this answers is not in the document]
hollandjg: So how would the table trajectories look afterwards?

like this?

```julia
julia> trajectories[:, [:ID, :passtime, :area_mismatch, :corr]]
6×4 DataFrame
 Row │ ID     passtime             area_mismatch  corr           
     │ Int64  DateTime             Float64?       Float64?       
─────┼───────────────────────────────────────────────────────────
   1 │     1  2022-09-14T12:00:00            0.0        0.991
   2 │     1  2022-09-16T12:00:00            0.0        0.995  
   3 │     1  2022-09-15T12:00:00      missing    missing      
   4 │     2  2022-09-14T12:00:00            0.0        0.988
   5 │     2  2022-09-16T13:00:00            0.0        0.988           
   6 │     2  2022-09-15T12:00:00      missing    missing

```

Could you alternatively change the meaning of `area_mismatch` and `corr` so that they mean the goodness of matching between the floe data in row `i` and the newest available earlier row `i-n`? Would that break other things?

test/test-long-tracker.jl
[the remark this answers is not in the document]
hollandjg: I think what you're getting with `maximum(ids)` is the ID of the trajectory with the highest ID, but not necessarily the count. 

You could get the count by counting the number of unique entries in `ids` (like `length(Set(ids))`). (I know that it's true right now, but I think the shortcut you've taken here obscures the intention behind your code, and if something breaks in future you or whoever is maintaining the code after you may have a difficult time understanding why you're using the functions you're using.)

test/test-long-tracker.jl
```diff
@@ -0,0 +1,79 @@
...
+        IDs = trajectories[!, :ID]
+        ids, counts = _imhist(IDs, unique(IDs))
+        @test maximum(ids) == 5
```
hollandjg, line 91: Oh, you're using two different `ids` and `IDs` variables which mean different things. Or are they the same?

test/test-long-tracker.jl
```diff
@@ -0,0 +1,79 @@
...
+
+        ids, counts = _imhist(counts, unique(counts))
```
hollandjg, line 93: And here you're taking a second histogram of the `counts` output from the line before. Why? I'm sure you can think of a more understandable way to write this.

src/tracker/tracker-funcs.jl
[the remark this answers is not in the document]
cpaniaguam: I think @danielmwatkins was fine with the current interpretation of those columns.

test/test-long-tracker.jl
cpaniaguam: The variable `IDs` refers to the data in the column of trajectory identification numbers `:ID`, which can have repeated values for trajectories of length greater than 1. The variable `ids` will contain the unique values in `IDs`. Feel free to make suggestions on how to improve the clarity of the code.

test/test-long-tracker.jl
cpaniaguam: > And here you're taking a second histogram of the counts output from the line before. Why?

The goal is to verify this claim: 
```
# Expected: 5 trajectories, 3 of which have length 3 and 2 of which have length 2
```
By using a histogram here, I can get a distribution of the id counts. Specifically, I am looking for 3 trajectories of length 3 (i.e., `3` repeated three times in counts) and 2 trajectories of length 2 (i.e., `2` appearing twice in counts). The histogram provides a clear way to verify this distribution. Could you please clarify what aspect of this approach you find unclear?

hollandjg: The use of similar names like `IDs` and `ids`, redefining `counts` to mean two different things in a few lines, and comparing `ids .== counts` are all confusing to me. This last line especially – `ids` is only equal to counts because of a quirk in the data, not because of anything fundamental.

cpaniaguam: Thank you for your feedback. Could you suggest specific changes to make the code clearer?

test/test-long-tracker.jl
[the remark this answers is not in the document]
cpaniaguam: > I think what you're getting with maximum(ids) is the ID of the trajectory with the highest ID, but not necessarily the count.

Please note that trajectories ids are enumerated `1:n` by design, so `maximun(ids)` is identical to `length(Set(ids))` or `length(unique(ids))`. In addition, building a histogram includes the workflow you are suggesting to get those unique ids. Again, feel free to make concrete code suggestion changes using the _suggestion_ feature. Thanks.

hollandjg (approved):

PR #532, opened 2024-12-19
Feat(long tracker): add separate UUID function

hollandjg:
Add function required for manipulating the props DataFrame to include a UUID column. 

This is required for the long-tracker CLI. In the "old" tracker `IceFloeTracker.pairfloes`, this step was included, but in the "long" tracker `IceFloeTraker.long_tracker`, it isn't, so it needs to be made available to go into the `long_tracker` version of the CLI wrapper script.

### DIFF
--- a/src/tracker/long_tracker.jl
+++ b/src/tracker/long_tracker.jl
@@ -22,7 +22,6 @@ Trajectories are built in two steps:
     - "orientation"
     - "perimeter"
     - "mask": 2D array of booleans
-    - "ID": unique identifier
     - "passtime": A timestamp for the floe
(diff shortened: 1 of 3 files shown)


src/tracker/long_tracker.jl
```diff
@@ -25,6 +25,7 @@ Trajectories are built in two steps:
     - "ID": unique identifier
```
cpaniaguam, line 25:
```suggestion
```

cpaniaguam (changes requested):
All good here, but consider cleaning up one of the docstrings.

cpaniaguam (approved):

Issue #541, opened 2025-01-24
Update package authors in project.toml


cpaniaguam:
What should it look like?
@danielmwatkins @hollandjg  @tdivoll

cpaniaguam:
Completed in #543

PR #543, opened 2025-01-24
chore: update version to 0.6.3

cpaniaguam:


Project.toml
```diff
@@ -1,7 +1,14 @@
 name = "IceFloeTracker"
 uuid = "04643c7a-9ac6-48c5-822f-2704f9e70bd3"
-authors = ["bradford roarr <bradford_roarr@brown.edu> and contributors"]
-version = "0.6.2"
+authors = [
+    "Carlos Paniagua <carlos_paniagua@brown.edu>",
+    "Timothy Divoll <bradford_roarr@brown.edu>",
+    "John Gerrard Holland <john_holland1@brown.edu>",
+    "bradford roarr <bradford_roarr@brown.edu>",
```
hollandjg, line 7: Name-case for Brad's name?
```suggestion
    "Bradford Roarr <bradford_roarr@brown.edu>",
```

hollandjg (changes requested):
There's one change I'd suggest

Project.toml
cpaniaguam: @broarr, any preference?

danielmwatkins: Good catch, definitely put all the names in the same format. I figure that if Brad had a super strong opinion about not using capital letters, it would be in lowercase on the CCV People page. I'd say capitalize it now and we can fix the format later if there's a different form of the name wanted.

cpaniaguam:
Closes #541

Project.toml
broarr: No preference on case on my name

hollandjg (approved):
Great!

PR #544, opened 2025-01-24
fix(tracker): reduce log size

hollandjg:
Get tracker working with new datasets
- reduce log size by removing some expensive debug statements which overwhelm log files if active
- allow `:label` column to promote from UInt8 to UInt16 (or higher) to allow for larger numbers of floes
- Add debug messages to `resolvecollisions!`


hollandjg:
I don't think this is still an issue – we've deactivated debugging messages in the pipeline – so I don't think we need to make these changes any more

PR #566, opened 2025-02-13
Unify small/large floe area thresholds

cpaniaguam:
Main change:

- Refactor condition_ thresholds from
```julia
search_thresholds = (dt=(30.0, 100.0, 1300.0), dist=(200, 250, 300))
large_floe_settings = (
    area=1200, # unify this
    arearatio=0.28,
    majaxisratio=0.10,
    minaxisratio=0.12,
    convexarearatio=0.14,
)
small_floe_settings = (
    area=1200, # and this to a common value
    arearatio=0.18,
    majaxisratio=0.1,
    minaxisratio=0.15,
    convexarearatio=0.2,
)
condition_thresholds = (search_thresholds, small_floe_settings, large_floe_settings)
```

to 
```julia
search_thresholds = (dt=_dt, dist=_dist)
large_floe_settings = (
    arearatio=0.28, majaxisratio=0.10, minaxisratio=0.12, convexarearatio=0.14
)
small_floe_settings = (
    arearatio=0.18, majaxisratio=0.1, minaxisratio=0.15, convexarearatio=0.2
)
large_floe_minimum_area = 1200 # single common value
condition_thresholds = (
    large_floe_minimum_area=large_floe_minimum_area, # new common value
    search_thresholds=search_thresholds,
    small_floe_settings=small_floe_settings,
    large_floe_settings=large_floe_settings,
)
```


test/test-long-tracker.jl
```diff
@@ -56,16 +32,18 @@ end
 begin # Filter out floes with area less than `floe_area_threshold` pixels
     floe_area_threshold = 400
     for (i, prop) in enumerate(_props)
-        _props[i] = prop[prop[:, :area].>=floe_area_threshold, :] # 500 working good
-        sort!(_props[i], :area, rev=true)
+        _props[i] = prop[prop[:, :area] .>= floe_area_threshold, :] # 500 working good
+        sort!(_props[i], :area; rev=true)
```
hollandjg, line 36: this line seems to me not to be a trivial change as it potentially switches which `sort!` you're using. Is that intentional? Is it fixing a bug? I found something similar which I had to patch a few months ago.

src/tracker/tracker-funcs.jl
```diff
@@ -221,23 +221,25 @@ end
...
     area1,
     ratios,
+    large_floe_minimum_area = 1200,
     large_floe_settings=(
-        area=1200, arearatio=0.28, majaxisratio=0.10, minaxisratio=0.12, convex_area=0.14
+        arearatio=0.28, majaxisratio=0.10, minaxisratio=0.12, convex_area=0.14
     ),
 )
-    return area1 > large_floe_settings.area &&
+    return area1 > large_floe_minimum_area &&
```
hollandjg, line 242: Shouldn't this be 
```suggestion
    return area1 >= large_floe_minimum_area &&
```
and the other be `<`? If I'm at the minimum area, then I fulfill the condition. If I'm smaller than it, I don't. So This should be greater than or equal to.

hollandjg (commented):
I'm not sure I understand the motivation of the refactoring. Can you say a bit more about why the new structure of the `condition_thresholds` is better? Is it just that you've converted it into a named-tuple and added a common parameter?

And I was primed to see the minimum floe area as an option here, but it looks like you've not included that in this PR. Is that correct?

cpaniaguam:
> I'm not sure I understand the motivation of the refactoring. Can you say a bit more about why the new structure of the `condition_thresholds` is better? Is it just that you've converted it into a named-tuple and added a common parameter?

See linked issue #566.

test/test-long-tracker.jl
cpaniaguam: As noted elsewhere, this will be removed when merging #570.

src/tracker/tracker-funcs.jl
cpaniaguam: 👍

test/test-long-tracker.jl
cpaniaguam: Also, this is a change in formatting; the two are equivalent.

hollandjg (dismissed):
Looks good to me!

hollandjg (approved):
Still looks good.

PR #570, opened 2025-02-13
Add control for small floes in tracker

cpaniaguam:
This pull request includes changes to the floe tracking algorithm in the `track-floes` notebook, the `long_tracker` function, and the corresponding test file. The main updates involve the handling of floes based on their area.

### Changes to floe area handling:
* [`src/tracker/long_tracker.jl`](diffhunk://#diff-c2a44906d1ee409a282e1c737d8d655be8ab0417c103a891e7eda6235ece547aR30-R41): Added a new parameter `small_floe_minimum_area` to the `long_tracker` function to filter out floes with an area less than this value. The default is set to 400 pixels.
* [`test/test-long-tracker.jl`](diffhunk://#diff-9a101e482af868fa2ddf398664d27679cb0a7951d087435ac0832c3afcf0099bL32-L39): Removed the filtering of floes with an area less than a specified threshold in the test setup.
* [`notebooks/track-floes/track-floes.ipynb`](diffhunk://#diff-11ee35f28376219715471e79af5315f2278f5b25347ff2e62dd58786af7a6c40L101-L108): Removed the filtering of floes with an area less than a specified threshold.


src/tracker/long_tracker.jl
```diff
@@ -27,11 +27,23 @@ Trajectories are built in two steps:
...
 
 # Returns
 A DataFrame with the above columns, plus two extra columns, "area_mismatch" and "corr", which are the area mismatch and correlation between a floe and the one that follows it in the trajectory. Trajectories are identified by a unique identifier, "uuid".
 """
-function long_tracker(props::Vector{DataFrame}, condition_thresholds, mc_thresholds)
+function long_tracker(
+    props::Vector{DataFrame},
+    condition_thresholds,
+    mc_thresholds,
+    small_floe_minimum_area=400,
```
hollandjg, line 39: Didn't we say we were going to include this in `condition_thresholds` like the other minimum area threshold?

src/tracker/long_tracker.jl
```diff
@@ -27,11 +27,23 @@ Trajectories are built in two steps:
...
+)
+    begin # Filter out floes with area less than `small_floe_minimum_area` pixels
+        for (i, prop) in enumerate(props)
+            props[i] = prop[prop[:, :area] .>= small_floe_minimum_area, :]
+            sort!(props[i], :area; rev=true)
+        end
+    end
```
hollandjg, line 46: This bit looks fine to me.

src/tracker/long_tracker.jl
cpaniaguam: It really doesn't belong there. The data in `condition_thresholds` is used in a unrelated task.

hollandjg: I think it's used in a different function, but it's a related task: Isn't it the same kind of thing as the `large_floe_minimum_area`?

cpaniaguam: I can put in the tuple but I don't think they are related. `small_floe_minimum_area` is intended to remove floes that are too small for tracking. `large_floe_minimum_area` is used to partition the area values which _can be tracked_ into "small" and "large". @danielmwatkins

cpaniaguam: The names can lead one to think they are. Maybe alternate names should be chosen?

hollandjg: I agree that they are being used in different functions right now, but I think that putting them together makes more sense from an interface perspective. But I could imagine a situation where we upgrade the tracker to deal with tiny floes too, and I think that taking that long view would be smarter. 

We might include all the floes detected (so undo the filtering), and then you'd have a three step condition:
- First look at tiny and small floes `0 <= area < large_floe_settings.minimum_area` and get the vector field describing floe movements from cross-correlation
- Then look at small floes `small_floe_settings.minimum_area <= area < large_floe_settings.minimum_area` and calculate the condition for those
- Then look at large floes `large_floe_settings.minimum_area <= area` and calculate the condition for those

Perhaps putting them in here:
```
small_floe_settings = (
    minimum_area=300, 
    arearatio=0.18, 
    majaxisratio=0.1, 
    minaxisratio=0.15, 
    convexarearatio=0.2
)
large_floe_settings = (
    minimum_area=1200, 
    arearatio=0.28, 
    majaxisratio=0.10, 
    minaxisratio=0.12, 
    convexarearatio=0.14
)
condition_thresholds = (;
    search_thresholds,
    small_floe_settings,
    large_floe_settings,
)
```

danielmwatkins: I also think that it makes sense to have it included with the floe size settings like John described. For now, that would make it so the user can interpret the "minimum area" in a similar way for both:
- If the floe is smaller than `small_floe_minimum_area`, don't track it
- If larger than `small_floe_minimum_area` and less than `large_floe_minimum_area`, track it with the small floe settings
- If larger than the `large_floe_minimum_area`, track it with the large floe settings.
My intuition tells me that based on the discrete geometry and shape statistics work that I'm doing that the transition points will be definable based on object to pixel size ratios and the associated uncertainty in shape properties. In remote sensing there's a division between low, moderate, and high resolution depending on the object-pixel size ratio, and I think a similar principle will apply here.

src/tracker/tracker-funcs.jl
```diff
@@ -250,30 +246,20 @@ end
...
-        arearatio=0.18,
-        majaxisratio=0.07,
-        minaxisratio=0.08,
-        convexarearatio=0.09,
-    ),
+    thresholds
 )
+    small_floe_settings = thresholds.small_floe_settings
+    large_floe_minimum_area = thresholds.large_floe_settings.minimum_area
     return area1 < large_floe_minimum_area &&
```
hollandjg:
```suggestion
    large_floe_settings = thresholds.large_floe_settings
    return area1 < large_floe_settings.minimum_area &&
```

src/tracker/long_tracker.jl
```diff
@@ -161,18 +168,28 @@ function find_floe_matches(
...
+    minimum_area=large_floe_minimum_area,
+    arearatio=0.28,
+    majaxisratio=0.10,
+    minaxisratio=0.12,
+    convexarearatio=0.14,
 )
+
 small_floe_settings = (
-    arearatio=0.18, majaxisratio=0.1, minaxisratio=0.15, convexarearatio=0.2
+    minimum_area=small_floe_minimum_area,
```
hollandjg: for consistency with the other names, you might want the name in this tuple to be `minimumarea` without the underscore.

hollandjg (approved):
Looks good to me!

PR #596, opened 2025-03-18
595 mismatch fails to find negative and large rotation rates

cpaniaguam:
This pull request introduces a new test script for evaluating the rotation and rescaling of ice floe shapes, and makes a small adjustment to the `mismatch` function in the `register-mismatch.jl` file. The most important changes include the addition of the test script and the modification of the rotation calculation in the `mismatch` function.

New test script for ice floe shapes:

* [`rotation_tests/test_floe_shapes.jl`](diffhunk://#diff-d23768c8c6efb6a1198eaa58ed2801c50e66893687e60d4df051f19e87650e15R1-R94): Added a comprehensive test script that loads labeled images, applies rotations and rescaling, and calculates various properties of the ice floes. The results are saved to CSV files.

Modification to rotation calculation:

* [`src/register-mismatch.jl`](diffhunk://#diff-3424903cdeb896799b12969d2f448388c6ab976528a58fb5603668c279847facL33-R34): Updated the `mismatch` function to use the `atan` function for calculating the rotation angle, improving the accuracy of the rotation estimation.


src/register-mismatch.jl
```diff
@@ -30,5 +30,6 @@ function mismatch(
         centered(fixed), centered(moving), mxshift, mxrot; print_interval=typemax(Int)
     )
 
-    return (mm=mm, rot=acosd(tfm.linear[1]))
+    θ = atand(tfm.linear[3] / tfm.linear[1])
```
hollandjg, line 33: You may need to change the way the `θ` is used later – is it compared with the threshold value directly, or do we take its absolute value first?

hollandjg (commented):
I think we'll need some additional tests of the `mismatch` function. I've started some in https://github.com/WilhelmusLab/ice-floe-tracker-pipeline/blob/feat(tracker)--add-rotation-calculations/IFTPipeline.jl/test/test-rotate.jl which we might be able to use.

src/register-mismatch.jl
cpaniaguam: Good point, @hollandjg. Although I was not intending to merge this PR, if angles can have negative values, this would need to be handled in calls to `matchcorr` where this function is actually used. Namely,

https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/e71e46c316bc1d5b491f90398258c58d4cacf2ec/src/tracker/matchcorr.jl#L54

So using the absolute value is necessary in the line above.

Having revisited this, I have recalled the motivation having set a default value for `mxrot` in calls to the `mismatch` function. In the MATLAB implementation, the estimated rotation is compared against a value of `10`. 

https://github.com/WilhelmusLab/ice-floe-tracker/blob/0076a08836b2de271f254bd2a6e3cc852acaf5f1/existing_code/input/info/match_corr.m#L666

That value generalized into a parameter for the tracker workflow. So I thought a default value `π/4` was "reasonable" for the purposes of passing it to `mismatch` as it is a required argument of `qd_rigid` and keep argument names the same across the different functions in the ecosystem.

danielmwatkins: Definitely a reasonable default value. 

Setting a maximum rotation is one of the things that needs to be justified one way or another. I'm still hoping to find some cases of observations that let me make more justified upper bounds. For now, it seems like the quadDIRECT algorithm has some limitations that we're running into. So having a warning at least in the documentation that the qd_rigid function will fail to find rotations higher than 67.5 is important. 

As with the travel distance parameterization, I think an issue we're running into is the way that the MATLAB code used step functions where continuous functions would make more sense and be less complicated to code. By default we're saying a 10 degree rotation is OK for a gap of 1.5 hours -- _and_ it's the limit for 24 hours or 48 hours. If something is moving at 10 degrees / 1.5 hours, that's 160 degrees in a day! 

There's not much data in the literature that I've found so far on ice floe rotation rates. However anecdotally, a 2 km - across floe with the MOSAiC camp on it did 360 degree rotations in the Fram Strait. While this doesn't prove that it's common, it does show that very high rotation rates can occur.

hollandjg:
resolved by #608

PR #599, opened 2025-03-24
chore: remove pyproj and rasterio dependencies

hollandjg:
Remove latlon.py and the dependencies associated with it.

- [x] Remove all the python `getlatlon` calls and replace with Julia `latlon`
- [x] Change all the places the data from `getlatlon` are used to use the new syntax


cpaniaguam (approved):
lgtm

PR #606, opened 2025-04-07
switch registration algorithm to naïve

### LINKED ISSUE #595: `mismatch` fails to find negative and large rotation rates
I set up a simple experiment to test the uncertainty in estimated orientation rate, taking observed floe objects and rotating them from -90 to 90 in steps of a couple of degrees.  It seems that the `mismatch` function only can return values in the range 0 to 67.5. The error is small in this range, but large outside of it. Note that without measurement data, we cannot assume any range of rotation out of hand, so we need to be able to handle large rotations. And of course we can't assume that the rotation will always be counterclockwise. 

Image

I suspect that the error is in casting the output of the quadDIRECT algorithm into degrees using `acosd`, but I could be wrong.

The "scale" parameter is the size of the floe: I resized objects to test the uncertainty as a function of floe area. Thus you can see that the smaller floes have more deviation from the 1-1 line as we might expect.

hollandjg:
Use @danielmwatkins's new symmetric difference mismatch minimization.

- reverses #294 
- resolves #595


test/test-register.jl
```diff
@@ -0,0 +1,2046 @@
...
+                    )
+                    absolute_error_wrt_all_aliased_angles = [
+                        abs(oriented_angle_between_angles(Δθ + offset, Δθ_measured)) for
+                        offset in angle_aliases
+                    ]
+                    minimum_absolute_error = minimum(absolute_error_wrt_all_aliased_angles)
+                    ok = minimum_absolute_error < deg2rad(precision_goal_degrees)
+                    push!(results, (; θ1, θ2, Δθ, Δθ_measured, minimum_absolute_error, ok))
+                end
+            end
```
cpaniaguam, line 35:
```suggestion
            for (θ1, mask1) in masks, (θ2, mask2) in masks
                Δθ = oriented_angle_between_angles(deg2rad(θ1), deg2rad(θ2))
                Δθ_measured = register(
                    mask1,
                    mask2;
                    imrotate_function=imrotate_bin_counterclockwise_radians
                )
                absolute_error_wrt_all_aliased_angles = [
                    abs(oriented_angle_between_angles(Δθ + offset, Δθ_measured)) for
                    offset in angle_aliases
                ]
                minimum_absolute_error = minimum(absolute_error_wrt_all_aliased_angles)
                ok = minimum_absolute_error < deg2rad(precision_goal_degrees)
                push!(results, (; θ1, θ2, Δθ, Δθ_measured, minimum_absolute_error, ok))
            end
```

test/test-register.jl
```diff
@@ -0,0 +1,2046 @@
...
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 1 1 1 1 1 0 0 0 0 0
+                    0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
+                    0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
+                ],
+            )
```
cpaniaguam, line 349: The test data is somewhat lengthy. Could they live in a separate file? This would streamline the test itself and keep its logic focused on the core functionality.

hollandjg: Yes, it's lengthy. But I think that having it in a single file makes comprehending what is happening that much easier. There's no having to look in a different file for the object, and trying to work out whether you're looking at the right array – they are right there. Putting them in a separate file would entail having to import them, potentially dealing with errors due to that. I tend to just collapse those Dictionaries using code folding in my editor, and then the file becomes more manageable to look at.

src/register.jl
```diff
@@ -0,0 +1,142 @@
...
+"""
+Pad images by zeros based on the size of the larger of the two images.
+"""
+function pad_images(im1, im2)
+    max1 = maximum(size(im1))
+    max2 = maximum(size(im2))
+
+    n = Int64(ceil(maximum([max1, max2])))
+    im1_padded = collect(padarray(im1, Fill(0, (n, n), (n, n))))
+    im2_padded = collect(padarray(im2, Fill(0, (n, n), (n, n))))
```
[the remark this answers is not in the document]
hollandjg: What comprehension would you suggest?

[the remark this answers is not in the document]
hollandjg: Like this?
```julia
im_padded = [collect(padarray(im, Fill(0, (n, n), (n, n)))) for im in [im1, im2]]
```

test/test-register.jl
hollandjg: Sure!

src/register.jl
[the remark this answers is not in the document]
cpaniaguam: Right!

src/register.jl
```diff
@@ -0,0 +1,142 @@
...
+
+# Functions used for the SD minimization
+"""
+Pad images by zeros based on the size of the larger of the two images.
+"""
+function pad_images(im1, im2)
+    max1 = maximum(size(im1))
+    max2 = maximum(size(im2))
+
+    n = Int64(ceil(maximum([max1, max2])))
```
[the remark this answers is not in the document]
hollandjg: Yes, we can simplify that a bit.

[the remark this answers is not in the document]
hollandjg: I've reduced the number of function calls to 3, with two tuple expansions

hollandjg:
Added some more tests of the padding.

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+    R = sum(im .> 0)
+    n, m = size(im)
+    for ii in range(1, n)
+        for jj in range(1, m)
+            if im[ii, jj] > 0
+                xi += ii
+                yi += jj
+            end
+        end
+    end
```
cpaniaguam, line 41: One can avoid nested for loops with `eachindex` or `CartesianIndices`. 
```suggestion
    for idx in CartesianIndices(im)
        if im[idx] > 0
            ii, jj = Tuple(idx)
            xi += ii
            yi += jj
        end
    end
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+
+    x0, y0 = sum(xi) / R, sum(yi) / R
+    if rounded
+        return round(Int32, x0), round(Int32, y0)
+    else
+        return x0, y0
+    end
```
cpaniaguam, line 48: This pattern is more Julian:
```suggestion
    rounded && return round(Int32, x0), round(Int32, y0)
    return x0, y0
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+should already be padded so that there is no danger of cutting into the floe shape.
+"""
+function crop_to_shared_centroid(im1, im2)
+    r1, c1 = compute_centroid(im1; rounded=true)
+    r2, c2 = compute_centroid(im2; rounded=true)
+
+    n1, m1 = size(im1)
+    n2, m2 = size(im2)
+    new_halfn = minimum([minimum([r1, n1 - r1]), minimum([r2, n2 - r2])])
+    new_halfm = minimum([minimum([c1, m1 - c1]), minimum([c2, m2 - c2])])
```
cpaniaguam, line 63: Consider using broadcasting:
```suggestion
    new_halfn = minimum(minimum.([r1, n1 - r1], [r2, n2 - r2]))
    new_halfm = minimum(minimum.([c1, m1 - c1], [c2, m2 - c2]))
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+Use `imrotate_function=imrotate_bin_<clockwise|counterclockwise>_<radians|degrees>` to get angles <clockwise|counterclockwise> in <radians|degrees>.
+"""
+function shape_difference_rotation(im_reference, im_target, test_angles; imrotate_function=imrotate_bin_clockwise_radians)
+    imref_padded, imtarget_padded = pad_images(im_reference, im_target)
+    shape_differences = Array{
+        NamedTuple{(:angle, :shape_difference),Tuple{Float64,Float64}}
+    }(
+        undef, length(test_angles)
+    )
+    # shape_differences = zeros((length(test_angles), 2))
```
cpaniaguam, line 91:
```suggestion
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+    # r_init, c_init = compute_centroid(imref_padded, rounded=true)
+    for angle in test_angles
+
+        # rotate image back by angle
+        imtarget_rotated = imrotate_function(imtarget_padded, -angle)
+
+        im1, im2 = crop_to_shared_centroid(imref_padded, imtarget_rotated)
+
+        # Check here that im1 and im2 sizes are the same
+        if isequal.(prod(size(im1)), prod(size(im2)))
```
cpaniaguam, line 103:
```suggestion
        if isequal(prod(size(im1)), prod(size(im2)))
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+            a_not_b = im1 .> 0 .&& isequal.(im2, 0)
+            b_not_a = im2 .> 0 .&& isequal.(im1, 0)
```
cpaniaguam, line 105: If all operations are elementwise you can use the `@.` macro.
```suggestion
            @. a_not_b = (im1 > 0) && (im2 == 0)
            @. b_not_a = (im2 > 0) && (im1 == 0)
```

src/register.jl
```diff
@@ -0,0 +1,138 @@
...
+The default test angles are evenly distributed in steps of π/36 rad (5º) around a full rotation,
+ensuring that no angles are repeated (since -π rad == π rad),
+and ordered so that smaller absolute angles which are positive will be returned in the event of a tie in the shape difference.
+Use `imrotate_function=imrotate_bin_<clockwise|counterclockwise>_<radians|degrees>` to get angles <clockwise|counterclockwise> in <radians|degrees>.
+"""
+
+function register(
+    im_reference,
+    im_target;
+    test_angles=sort(reverse(range(; start=-π, stop=π, step=π / 36)[1:(end-1)]); by=abs),
```
cpaniaguam, line 130: I think it'd be good to define this constant elsewhere before the function's definition. It might even be better to define it within the function as a function of the number of steps.

test/test-register.jl
```diff
@@ -0,0 +1,2044 @@
+using Dates
+using LinearAlgebra: dot, det, norm
+using IceFloeTracker: register, imrotate_bin_counterclockwise_radians
+
+@testset "rotation" begin
+    @testset "individual rotations" begin
+        unit_vector(θ) = [cos(θ); sin(θ)]
+        oriented_angle_between_vectors(a, b) = atan(det(hcat(a, b)), dot(a, b))
```
cpaniaguam, line 8: Didn't we have this change in already?
```suggestion
        oriented_angle_between_vectors(u, v) = atan(det(hcat(u, v)), dot(u, v))
```

src/register.jl
hollandjg: I think we need some direct test cases of this function before we start modifying the functions – this is an okay change, but I don't think we have the testing yet for refactoring.

cpaniaguam: It's a different way to do the same work. Would this suffice?
```julia
julia> old(r1, n1, r2, n2) = minimum([minimum([r1, n1 - r1]), minimum([r2, n2 - r2])])
old (generic function with 1 method)

julia> new(r1, n1, r2, n2) = minimum(minimum.([[r1, n1 - r1], [r2, n2 - r2]]))
new (generic function with 1 method)

julia> r1, n1, r2, n2 = rand(1:100_000, 4);

julia> @assert old(r1, n1, r2, n2) == new(r1, n1, r2, n2)

julia>

Comment (hollandjg) [reply src/register.jl:61]: I mean more tests of the `crop_to_shared_centroid` – just to be sure we're not inadvertantly introducing a regression when we change things.

**Comment by hollandjg** [comment]
I'm adding tests of the centroid, cropping, and mismatch functions before we start refactoring any more code. Shouldn't change behavior, but I want to avoid introducing any regressions.

File: src/register.jl (hunk 7 above)
Comment (cpaniaguam) [reply src/register.jl:61]:
```julia
julia> @assert old(r1, n1, r2, n2) == new(r1, n1, r2, n2) == minimum([r1, n1 - r1, r2, n2 - r2])

julia>

File: src/register.jl (hunk 10 above)
Comment (hollandjg) [reply src/register.jl:102]: This causes a crash unless `a_not_b` and `b_not_a` are already defined. I think it's easier to just leave them as they are.

Comment (cpaniaguam) [reply src/register.jl:104]: I see. The `@.` should be to the right of the assignment operator otherwise it'd be converted to `.=` also.
```julia
julia> im1, im2 = [rand(0:1, 10,10) for _ in 1:2];

julia> foo = @. (im1 > 0) && (im2 == 0)

julia>

File: src/register.jl (hunk 11 above)
Comment (hollandjg) [reply src/register.jl:129]: I don't want to define it as a number of steps because we have cases where you might want to check the best fit for some non-continuous range of angles. 

For instance, if you have a rotational symmetry of order 2, and you know that you have a strong semimajor-axis measurement along the x-axis, you might want to look for the best fit in the angles `[0., 0+π/10, 0-π/10, 9π/10, 11π/10]` but you wouldn't want to look between those.

Open to including this as a variable though.

**Comment by hollandjg** [comment]
I hit a bug with the handling of pairs of floe images where one had an odd number of rows/columns and the other had an even number. I've updated the alignment algorithm with one which handles both of those cases nicely, and dropped some now-unused code.

File: src/register.jl [hunk 13]
```diff
@@ -0,0 +1,166 @@
...
+Use `imrotate_function=imrotate_bin_<clockwise|counterclockwise>_<radians|degrees>` to get angles <clockwise|counterclockwise> in <radians|degrees>.
+"""
+function register(
+    im_reference,
+    im_target;
+    test_angles=register_default_angles_rad,
+    imrotate_function=imrotate_bin_clockwise_radians,
+)
+    shape_differences = shape_difference_rotation(im_reference, im_target, test_angles; imrotate_function)
+    best_match = argmin((x) -> x.shape_difference, shape_differences)
```
danielmwatkins, line 126: Is it possible that this will yield multiple matches?

src/register.jl
```diff
@@ -0,0 +1,166 @@
...
+    return best_match.angle
+end
+
+"""
+The default registration angles are evenly distributed in steps of 5º around a full rotation,
+ensuring that no angles are repeated (since -180º == -180º),
+and ordered so that smaller absolute angles which are positive will be returned in the event of a tie in the shape difference.
```
danielmwatkins, line 133: Oh okay - so argmin defaults to just returning the first instance?

danielmwatkins (approved):
I think this looks good. Glad you found something that worked for the even/odd centering issue.

src/register.jl
hollandjg: Yes:
```julia
julia> argmin((x)->x.v, [(;id=1, v=1), (;id=2, v=1), (;id=3, v=1)])
(id = 1, v = 1)
```

PR #612, opened 2025-04-29
Changed filenames for accuracy

danielmwatkins:
The test filenames with "NE_Greenland" were incorrectly named -- the images show the Beaufort and Chukchi Seas, Greenland is not visible in the images. Furthermore, "reflectance" was used instead of "falsecolor" -- both truecolor and falsecolor images are corrected reflectance, so it doesn't make sense to use the word "reflectance" here.

I've not changed the filenames in example notebooks, I'll handle that in a separate pull request as I'm updating the notebooks today.

### DIFF
--- a/notebooks/preprocessing-workflow/preprocessing-workflow.ipynb
+++ b/notebooks/preprocessing-workflow/preprocessing-workflow.ipynb
@@ -161,11 +161,11 @@
    "source": [
     "# 1. Load test data\n",
     "\n",
-    "reflectance_test_image_file = \"test_inputs/NE_Greenland_reflectance.2020162.aqua.250m.tiff\"\n",
+    "reflectance_test_image_file = \"test_inputs/beaufort-chukchi-seas_falsecolor.2020162.aqua.250m.tiff\"\n",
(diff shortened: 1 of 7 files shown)


danielmwatkins:
Hi @cpaniaguam , I'm a bit confused by the errors that I'm getting in the check notebooks tests. Can you take a look? It looks to me like it's throwing a file not found error for the NE_greenland image during the track-floes test, but that notebook doesn't explicitly load the NE_greenland image. And I changed the inputs in the preprocess-workflow file already.

cpaniaguam:
> Hi @cpaniaguam , I'm a bit confused by the errors that I'm getting in the check notebooks tests. Can you take a look? It looks to me like it's throwing a file not found error for the NE_greenland image during the track-floes test, but that notebook doesn't explicitly load the NE_greenland image. And I changed the inputs in the preprocess-workflow file already.

@danielmwatkins 
There is one reference left when loading the image extracting lat/lon data under step 7.
```julia
latlonimgpth = joinpath(HOME, "test", "test_inputs", "NE_Greenland_truecolor.2020162.aqua.250m.tiff")
```

cpaniaguam (approved):
All good now!

PR #617, opened 2025-05-19
initial example notebook for cloud mask demo

danielmwatkins:
This pull request adds a notebook and a tiff image for demo-ing the cloud mask function. I've also add a temporary change to the cloudmask.jl functions to have a broader range of accepted image types; we should discuss in the issue on image types what the cleanest way to have a broad list of inputs would be.

A future TBD for the cloudmask notebook will be to show in a diagram how the cloud mask works, so that a user can be more comfortable with changing settings.


danielmwatkins:
Will rethink this with the updated API.

PR #625, opened 2025-06-05
WIP: simplified API

hollandjg:


src/cloudmask.jl
```diff
@@ -61,25 +18,34 @@ Convert a 3-channel false color reflectance image to a 1-channel binary matrix;
...
-    )
+    modis_721::AbstractArray{T};
+    prelim_threshold::Real=(110 / 255),
+    band_7_threshold::Real=(200 / 255),
+    band_2_threshold::Real=(190 / 255),
+    ratio_lower::Real=0.0,
+    ratio_upper::Real=0.75,
+)::BitMatrix where {T<:Union{AbstractRGB,TransparentRGB}}
+    modis_band02 = green.(modis_721)
+    modis_band07 = red.(modis_721)
```
danielmwatkins, line 29: I like that this approach makes it specific that this particular algorithm requires the MODIS 7-2-1 image type, and uses the green and red functions to get the channel info.

src/algorithms.jl
```diff
@@ -0,0 +1,234 @@
+abstract type IceFloeTrackerAlgorithm end
```
hollandjg, line 1: This should be an IceFloeSegementationAlgorithm

Would we need an IceFloeTrackerAlgorithm?
Would we need an IceFloeSegmentationTrackingAlgorithm?

hollandjg:
Convert to https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects notation

PR #637, opened 2025-07-02
feat: Add LopezAcosta2019Tiling() functor

hollandjg:
Add a functor for the tiling preprocessing algorithm. 

Includes a notebook showing the new functor in action: https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/d367fc17b4885df3aaac48e726e1f04492921ffa/notebooks/preprocessing-workflow/lopez-acosta-2019-tiling-workflow.ipynb

### DIFF
--- a/test/test-segmentation-lopez-acosta-2019-tiling.jl
+++ b/test/test-segmentation-lopez-acosta-2019-tiling.jl
@@ -118,5 +118,8 @@ using IceFloeTracker: LopezAcosta2019Tiling
             ".png",
             map(i -> segment_mean(segments, i), labels_map(segments)),
         )
+
+        foreground::BitMatrix = labels_map(segments) .> 0  # equivalent to the old segmentation
+        @test abs(sum(foreground) - 1461116) / 1461116 < 0.1
     end
 end


src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+
+    @info "Set ice labels thresholds"
+    ice_labels_thresholds = (
+        prelim_threshold=p.ice_labels_prelim_threshold,
+        band_7_threshold=p.ice_labels_band_7_threshold,
+        band_2_threshold=p.ice_labels_band_2_threshold,
+        ratio_lower=p.ice_labels_ratio_lower,
+        ratio_upper=p.ice_labels_ratio_upper,
+        r_offset=p.r_offset,
+    )
```
cpaniaguam, line 103: Why not pack these within `p`?

src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+    @debug ice_labels_thresholds
+
+    @info "Set adaptive histogram parameters"
+    adapthisteq_params = (
+        white_threshold=p.adapthisteq_white_threshold,
+        entropy_threshold=p.adapthisteq_entropy_threshold,
+        white_fraction_threshold=p.adapthisteq_white_fraction_threshold,
+    )
+    @debug adapthisteq_params
```
cpaniaguam, line 112: Pack these within `p`?

src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+
+    @info "Set gamma parameters"
+    adjust_gamma_params = (
+        gamma=p.gamma, gamma_factor=p.gamma_factor, gamma_threshold=p.gamma_threshold
+    )
```
cpaniaguam, line 117: Pack within `p`?

src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+    @info "Set structuring elements"
+    structuring_elements = p.structuring_elements
+    @debug structuring_elements
+
+    @info "Set unsharp mask params"
+    unsharp_mask_params = (
+        radius=p.unsharp_mask_radius,
+        amount=p.unsharp_mask_amount,
+        factor=p.unsharp_mask_factor,
+    )
```
cpaniaguam, line 129: Pack in?

src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+    @info "Set brighten factor"
+    brighten_factor = p.brighten_factor
+    @debug brighten_factor
+
+    @info "Set preliminary ice masks params"
+    prelim_icemask_params = (
+        radius=p.prelim_icemask_radius,
+        amount=p.prelim_icemask_amount,
+        factor=p.prelim_icemask_factor,
+    )
```
cpaniaguam, line 141: Pack?

src/preprocess_tiling.jl
```diff
@@ -1,93 +1,166 @@
...
+    ice_masks_params = (
+        band_7_threshold=p.icemask_band_7_threshold,
+        band_2_threshold=p.icemask_band_2_threshold,
+        band_1_threshold=p.icemask_band_1_threshold,
+        band_7_threshold_relaxed=p.icemask_band_7_threshold_relaxed,
+        band_1_threshold_relaxed=p.icemask_band_1_threshold_relaxed,
+        possible_ice_threshold=p.icemask_possible_ice_threshold,
+        k=p.icemask_n_clusters, # number of clusters for kmeans segmentation
+        factor=255, # normalization factor to convert images to uint8
+    )
```
cpaniaguam, line 154: Pack?

cpaniaguam (changes requested):
Thanks for this. Consider creating the tuples within the stuct as these values are a part of what defines the algorithm.

src/preprocess_tiling.jl
hollandjg: I'll think about how we can make that happen in a way which still makes it simple to modify single parameters when running. Might have to be nested structs.

hollandjg: I think one would ideally want the instantiation of the algorithm in that case to look like:
```
p = LopezAcosta2019(gamma=(threshold=250))
```
... which would leave the other gamma parameters as their default values:
```
p.gamma = (gamma=1.5, factor=1.3, threshold=250)

I don't see an obvious way to do that with pure NamedTuples.

You could sort-of do that with Structs, but you might have to write it like this:
```
p = LopezAcosta2019(gamma=GammaCorrectionAlgorithm(threshold=250))
```
... which I find somewhat unwieldy and requires the user to have a deeper understanding of the nuts-and-bolts of the algorithm than I think they should need.

If you've any ideas of how to make the interface user friendly and still be easy to code and understand in the code, please let me know.

Comment (danielmwatkins) [reply src/preprocess_tiling.jl]: Carlos, is your idea to have the tuples be arguments so you would specify a set of thresholds for each subroutine?

Like so:
```
examp_ice_labels = (prelim_threshold=100,
        band_7_threshold=200,
        band_2_threshold=110,
        ratio_lower=0,
        ratio_upper=0.75,
        r_offset=0)

p = LopezAcosta2019(ice_labels_thresholds = examp_ice_labels)
```

hollandjg:
Superseded by https://github.com/WilhelmusLab/IceFloeTracker.jl/pull/670

PR #639, opened 2025-07-03
Loosen package compatibility restrictions

hollandjg:


Project.toml
```diff
@@ -37,33 +37,35 @@ Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
...
+OffsetArrays = ">= 1.12"
+Peaks = ">= 0.5"
+Proj = ">= 1.7.2"
+PyCall = ">= 1.95"
+StaticArrays = ">= 1.5"
+StatsBase = ">= 0.34"
+TOML = ">= 1"
+TiffImages = "^0.10.0"
+TiledIteration = ">= 0.3"
+TimeZones = ">= 1.21.3"
```
cpaniaguam, line 68: I think adding `TiffImages` might be all that's needed. Using version lower bounds with `>=` will potentially install breaking versions of packages.

hollandjg: I don't know why the maintainers of those other packages do what they do. 

Some do include ranges; from Images.jl:
```
[compat]
...
FFTW = "1.5, 1.6, 1.7, 1.8"
...
ImageBinarization = "0.2, 0.3"
```
... which we have mostly not been doing up until now. 

In the case where the package "bundles" and re-exports some other packages, such as [Images.jl](https://github.com/JuliaImages/Images.jl/blob/master/Project.toml):
```
...
ImageAxes = "0.6.9"
ImageBase = "0.1.5"
ImageBinarization = "0.2, 0.3"
ImageContrastAdjustment = "0.3.12"
ImageCore = "0.9.3, 0.10"
ImageCorners = "0.1.2"
ImageDistances = "0.2.5"
ImageFiltering = "0.7.1"
ImageIO = "0.5, 0.6"
ImageMagick = "1.4"
ImageMetadata = "0.9.7"
ImageMorphology = "0.4"
ImageQualityIndexes = "0.3"
ImageSegmentation = "1.7"
ImageShow = "0.3"
...
```
... I think you can treat the parent package `Images.jl` as being the one you specify, and let it handle the other transitive dependencies. Why they have the ImageBinarization and ImageCore dependency declared as a range I'm not sure. 

In our case, just as an example, by additionally restricting ImageCore to [0.9.3, 0.10.0) (non-inclusive upper bound) we disallowed the installation of the updated ImageSegmentation package which is supported by Images 0.26 but not by ImageCore 0.9. 

I'll go back and rewrite this PR to make smaller changes, drop some of our explicit dependencies.

PR #645, opened 2025-07-07
feat: add support for more image number formats to LopezAcosta2019()

hollandjg:
Add support for n0f8, n6f10, n4f12, n2f14, n0f16, float32, float64 image types. 

This PR shows a way we might want to run tests of all of the underlying functions – if each can run with all of these types, then we should be in good shape.


hollandjg:
I reverted to using the `save` function rather than the `@persist` macro here because the `@persist` wasn't compatible with string interpolations like `"./test_outputs/segmentation-Lopez-Acosta-2019-mean-labels_$(target_type).png",` – that would throw an error.

cpaniaguam:
>  I reverted to using the save function rather than the @persist macro here because the @persist wasn't compatible with string interpolations
Why not do the interpolation first and then call `@persist`?

danielmwatkins (approved):
Nice!

PR #652, opened 2025-07-09
refactor: tiling algorithm as functor

hollandjg:


danielmwatkins:
This looks like a nice way to conceptualize the tiling approach. Can you also update the tiling example notebook to demonstrate the method in action?

PR #657, opened 2025-07-10
Feat: Add uncertainty to tracker calculations

danielmwatkins:
* Correlation has a simple formula for a confidence interval, add this to the output of matchcorr
* Use validation dataset to estimate the uncertainty in the shape difference as a function of shape size
* Use the uncertainty in the shape difference to estimate the uncertainty in rotation

### DIFF
--- a/src/tracker/matchcorr.jl
+++ b/src/tracker/matchcorr.jl
@@ -22,8 +22,8 @@ A pair of `NaN` is returned for cases for which one of their mask dimension is t
 # Arguments
 - `f1`: mask of floe 1
 - `f2`: mask of floe 2
-- `Δt`: time difference between floes
-- `mxrot`: maximum rotation (in degrees) allowed between floesn (default: 10)
+- `Δt`: time difference between floes in minutes
+- `mxrot`: maximum rotation (in degrees) allowed between floes (default: 10)
 - `psi`: psi-s-correlation threshold (default: 0.95)
 - `sz`: size threshold (default: 16)
 - `comp`: size comparability threshold (default: 0.25)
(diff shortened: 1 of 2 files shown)


danielmwatkins:
A few other notes on things to update
1. In `matchcorr`, it looks to me like the function would compute correlation, then if the correlation was good, it set mm to 0 and returned, thus bypassing the mismatch calculation.
2. Even if it were to get to the last conditional statement, `matchcorr` has a hard limit of 300 for delta t. This regardless of settings in the long-tracker. 
3. If it got past all that and had mismatch less than 0.1, this was also erased.
4. The function `corr` is actually finding the maximum lagged correlation; we should avoid confusion with standard correlation and use a better name.

I'm looking at the notation, we have $c$ for correlation. Might be better to use $r$ or $\rho$. I'll need to update other places in the code where matchcorr is called to handle different lengths of returned named tuples.

hollandjg:
We could add something like:
```julia
function mismatch(fixed::AbstractArray, moving::AbstractArray; mxrot=180, step=1)
    test_angles = sort(
        reverse(range(; start=-mxrot, stop=mxrot, step=step)[1:(end - 1)]); by=abs
    )
    return mismatch(fixed, moving; test_angles=test_angles)
end
```
as an additional mismatch functions. We'll have to check how Julia handles the keyword arguments.

danielmwatkins:
I've updated the return values for the matchcorr function:
```
    (shape_difference=_mm,
    psi_s_correlation=r,
    rotation=rot,
    corr_ci=(rlow, rhigh),
    sd_ci=(sdlow, sdhigh,
    rotation_ci=(sdlow, sdhigh))
```

I need to decide how verbose to be for the confidence intervals; obviously it's inconsistent to use `sd_ci` and `rotation_ci`.

Next, I need to address the possibility of no minimum or multiple minima. The confidence interval may help: a confidence interval that covers most possible angles is an indicator that the minimum is uncertain. Alternatively, rot could be NaN if there minimum is not significant.

PR #661, opened 2025-07-30
Fix: update mismatch function to include mxrot

hollandjg:
Add version of mismatch function with an mxrot and step option; make that the default.


danielmwatkins (approved):
Looks good!

PR #663, opened 2025-07-30
test: add download dataset script

hollandjg:


danielmwatkins (commented):
This looks great and I think it will help making demonstration notebooks. Regarding organization, I think it would be useful to add a file to the src folder with segmentation metrics. I've been working on measuring segmentation performance for another paper, and there are a bunch of tools available from Python for doing that but not many that I've seen yet in Julia. So it would be useful to have the evaluation metrics in the main source code of the package.

PR #675, opened 2025-08-11
feat[LopezAcosta2019]: add intermediate results callback

hollandjg:
LopezAcosta2019 now accepts a function which will be called with all the intermediate results.

This can be used to store the results to file, or pass the results out to the caller, or anything else the user requires. It makes the return signature of the LopezAcosta2019 functor consistent – it always only returns the segmented image.

Includes an additional factory function, `callable_store`, which creates a data store dictionary and a callback function to write data to the dictionary, as an example of a utility which could export the intermediate data.

### DIFF
--- a/src/IceFloeTracker.jl
+++ b/src/IceFloeTracker.jl
@@ -49,7 +49,8 @@ export readdlm,
     ValidationDataLoader,
     Watkins2025GitHub,
     segmentation_comparison,
-    segmentation_summary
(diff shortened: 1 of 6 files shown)


danielmwatkins (approved):
This looks good to me. I like the approach of having the intermediate steps all be optional outputs. We'll need to update the demonstration notebooks when this is done so we can show how the individual functions come together into the functors.

PR #689, opened 2025-08-14
feat[find_ice_labels]: RGBA support

hollandjg:


danielmwatkins:
@hollandjg Is this superseded by the updated ice detection functors?

PR #690, opened 2025-08-15
WIP: refactor[find_ice_labels]: update landmask formats 

hollandjg:


test/test-find-ice-labels.jl
```diff
@@ -1,23 +1,169 @@
...
-        falsecolor_image[ice_floe_test_region...], landmask[ice_floe_test_region...]
-    )
+        @test binarize([water], f) == [is_not_ice]
+        @test binarize([sea_ice], f) == [is_ice]
+        @test binarize([grass], f) == [is_not_ice]
+        @test binarize([masked], f) == [is_not_ice]
+        @test binarize([shadow], f) == [is_not_ice]
+        @test binarize([land_ice], f) == [is_ice]
+        @test binarize([masked_land_ice], f) == [is_not_ice]
+    end
```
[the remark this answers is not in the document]
hollandjg: The thresholds stay the same. I chose the pixel values by looking at a couple of example images and using using the built-in color meter of macOS to select a pixel in one of the false-color reflectance images from each of the regions:

- deep dark water, 
- sea ice, grass, 
- transparency over land, 
- shadow of an icy hill on land, 
- ice on land, 
- and the same ice on land but with the alpha channel set to zero.

test/test-find-ice-labels.jl
```diff
@@ -1,23 +1,169 @@
...
 
-    DelimitedFiles.writedlm("ice_labels_floe_region.csv", ice_labels_ice_floe_region, ',')
+    @ntestset "IceDetectionBrightnessPeaksMODIS721" begin
+        f = IceDetectionBrightnessPeaksMODIS721(5 / 255, 75 / 255)
+
+        data_loader = Watkins2025GitHub(; ref="a451cd5e62a10309a9640fbbe6b32a236fcebc70")
+        case = first(data_loader(c -> (c.case_number == 12 && c.satellite == "terra")))
+
+        # There is only one pixel in this image marked as peak
```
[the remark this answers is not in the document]
hollandjg: The input image is this one: 
Image

The result of binarizing it using the brightness-peaks method as it currently is written is this:
Image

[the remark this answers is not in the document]
danielmwatkins: Ah, ok. I've noticed in other cases as well that the `find_ice` methods often end up selecting only a few specks. In the end the result can be fine if that speck happens to be inside a k-means segment. I'm working on testing the sensitivity to the threshold choices to see how much of a problem this is.

src/find_ice_labels.jl
```diff
@@ -1,37 +1,163 @@
...
+- `find_ice(image, a(; kwargs...))`
+- or `a(; kwargs...)(image)`.
 
 """
-    find_reflectance_peaks(reflectance_channel, possible_ice_threshold;)
+abstract type IceDetectionAlgorithm <: AbstractImageBinarizationAlgorithm end
 
-Find histogram peaks in single channels of a reflectance image and return the second greatest peak. If needed, edges can be returned as the first object from `build_histogram`. Similarly, peak values can be returned as the second object from `findmaxima`.
+function (a::IceDetectionAlgorithm)(image::AbstractArray{<:Colorant})
+    return binarize(image, a)
```
danielmwatkins, line 16: Very cool that the algorithm can now get passed to `binarize`!

src/find_ice_labels.jl
```diff
@@ -1,37 +1,163 @@
...
-}
+using ImageBinarization: AbstractImageBinarizationAlgorithm, binarize
+"""
+    IceDetectionAlgorithm
+
+Functors to detect ice regions in an image.
+
+Each algorithm `a` with parameters `kwargs...` can be called like:
+- `binarize(image, a(; kwargs...))` 
+- `find_ice(image, a(; kwargs...))`
```
danielmwatkins, line 9: The alias `find_ice` may be confusing if it only is for the binarization algorithms. Are you thinking of this as a function that could accept all the IFT Ice Detection Algorithms (even later ones that aren't binary?)

hollandjg: Honestly, I'm happy to get rid of it. It's just an alias for `binarize` after all. We can always add a different alias else back in later, if we need it. The assumption that the results will be binary are pretty deep in the code, so we'd need to do a lot more work to support a general non-binary algorithm.

src/find_ice_labels.jl
hollandjg: Yes, and we get all the other binarization algorithms for free, should we want them.

danielmwatkins:
So, I’ve been trying to see where the crash comes from. I found some unexpected results looking at histograms - for some images at least I’m seeing a noisy histogram where the find maxima is picking up tons of tiny maxima along with the peak. It may have to do with lossy conversion between image types. Either way I think I can get a cleaner estimate of the major peaks with a little care, so we don’t have just a fixed index lookup.-DanielOn Aug 25, 2025, at 17:36, John Gerrard Holland ***@***.***> wrote:﻿
@hollandjg commented on this pull request.



In src/find_ice_labels.jl:
>      _, counts = ImageContrastAdjustment.build_histogram(reflectance_channel)
     locs, _ = Peaks.findmaxima(counts)
     sort!(locs; rev=true)
-    return locs[2] # second greatest peak
+    return locs[2] / 255.0 # second greatest peak

Then the code crashes as before. We have at least three test cases in the current set which suffer from that issue. Do you have a preferred behavior in that case?

danielmwatkins (approved):
I like the new approach with masks and with the ice detection functors. Let's merge this, and then approach the peak detection function to make it more robust.

PR #694, opened 2025-08-25
fix[find_reflectance_peaks]: if no reflectance peaks are found, return an Inf value

hollandjg:
If no reflectance peaks, or if there are no valid pixels, return `Inf` rather than crashing for the peak value. This guarantees that nothing will be evaluated as larger than the result.

### DIFF
--- a/src/find_ice_labels.jl
+++ b/src/find_ice_labels.jl
@@ -96,11 +96,8 @@ function find_reflectance_peaks(
     _, counts = ImageContrastAdjustment.build_histogram(reflectance_channel)
     locs, _ = Peaks.findmaxima(counts)
     sort!(locs; rev=true)
-    if 2 ≤ length(locs)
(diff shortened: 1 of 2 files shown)


src/find_ice_labels.jl
```diff
@@ -96,7 +96,11 @@ function find_reflectance_peaks(
     _, counts = ImageContrastAdjustment.build_histogram(reflectance_channel)
     locs, _ = Peaks.findmaxima(counts)
     sort!(locs; rev=true)
-    return locs[2] / 255.0 # second greatest peak
+    if 2 ≤ length(locs)
+        return locs[2] / 255.0 # second greatest peak
+    else
+        return Inf
+    end
```
cpaniaguam, line 103:
```suggestion
    2 ≤ length(locs) && return locs[2] / 255.0 # second greatest peak
    return Inf
```

cpaniaguam (approved):

PR #702, opened 2025-08-27
test: reorganize to include e2e and smoke tests, runnable independently

hollandjg:
Flatten nested test-cases for the segmentation functors, so that we can label them and run them independently.

- Now when setting up devcontainer, only the "smoke" tests will run – some basic tests which ensure that all the parts are working, more or less, but don't take long.


danielmwatkins (approved):

danielmwatkins (approved):
Looks good to me!

PR #704, opened 2025-08-28
Feat: Simplify image histogram generation and analysis

danielmwatkins:
This pull request alters the image histogram functions in the `ice_masks`, `find_ice_labels`, and `histogram_equalization` functions. The custom histogram function `_imhist` is removed and instead the `build_histogram` function from Julia Images is used. 

To make the histogram analysis more robust, we replace index lookup with use of the edges vector. 

Components of pull request:
1. Replace `_imhist` usage in long-tracker tests by doing a groupby/count on the trajectories dataframe
2. Replace calls to imhist with build_histogram
3. Separate the histogram building from the identification of relevant image peaks in the `ice_masks` and `find_ice_labels` files.


hollandjg:
This is looking pretty good from the tests!
- Some of the skill measures have gone down a tiny bit in some of the tests, but not by much, so we could certainly loosen those tests. 
- One test shows a missing function name – `get_image_peaks`. I'm not sure yet whether it's being called by another part of the code, or if there's an explicit test of it. 

I've updated the `base` branch so the files changed screen only shows the things you've changed relative to the starting branch. I've also updated the code to include the fix to the notebooks error in #705.

danielmwatkins:
Thanks for taking a look at this, John! And for cleaning up the commit history by pointing it to the right branch.

The `get_image_peaks` function is in `ice_masks.jl`. Does it need to be added to the explicit "export" list in `src/IceFloeTracker.jl` to be accessible to the tests?

danielmwatkins:
This last commit has a few bigger changes in it. I've updated the get_nlabel function to use the IceDetectionFirstNonZeroAlgorithm approach that John put together for the non-tiling algorithm. I've taken the binarization out of the get ice labels function. And I've unified the two get reflectance peaks functions (get_image_peaks and _find_reflectance_peaks --> get_ice_peaks), with the renaming reflecting that there are multiple peaks in the image and we only want the one corresponding to sea ice. I've also taken out a lot of the spots where the image gets multiplied by 255 to turn it into pseudo uint8.

There are still some errors, from what I've seen so far it's all the "unexpected pass" variety which is potentially good.

src/find_ice_labels.jl
```diff
@@ -1,4 +1,30 @@
...
+    # Normalize the possible sea ice section of the histogram. 
+    # Images with a lot of masked pixels can have large peaks at 0, which
+    # we don't want to include in the normalization. If no potential
+    # ice pixels, then return early
+    counts = normalizer > 0 ? counts ./ normalizer : return Inf
+    pks = findmaxima(counts, window) |> peakproms! |> peakwidths!
+    pks_df = DataFrame(pks[Not(:data)])
+    pks_df = sort(pks_df, :proms, rev=true)
+    maximum(pks_df.proms) < minimum_prominence && return Inf
+    return edges[pks_df[argmax(pks_df.proms), :indices]]
```
cpaniaguam, line 25: You could use `findmax` here which produces both the max and argmax.
```suggestion
    mx, argmx = findmax(pks_df.proms)
    mx < minimum_prominence && return Inf
    return edges[pks_df[argmx, :indices]] 
```

src/find_ice_labels.jl
```diff
@@ -69,8 +95,13 @@ function (f::IceDetectionBrightnessPeaksMODIS721)(out, modis_721_image, args...;
...
 
-    band_2_peak = find_reflectance_peaks(band_2 .* alpha_binary; f.possible_ice_threshold)
-    band_1_peak = find_reflectance_peaks(band_1 .* alpha_binary; f.possible_ice_threshold)
+    band_2_peak = get_ice_peaks(
+                    build_histogram(band_2 .* alpha_binary, 64; minval=0, maxval=1)... ;
+                    possible_ice_threshold=f.possible_ice_threshold)
+
+    band_1_peak = get_ice_peaks(
+                    build_histogram(band_1 .* alpha_binary, 64; minval=0, maxval=1)... ;
+                    possible_ice_threshold=f.possible_ice_threshold)
```
cpaniaguam, line 104: If you like you could create a closure fixing all but the `band_x` argument. Something like this:
```julia
# Create the closure
get_band_peak = function(band)
    get_ice_peaks(
        build_histogram(band .* alpha_binary, 64; minval=0, maxval=1)...;
        possible_ice_threshold=f.possible_ice_threshold
    )
end

# Use the closure twice with different `band` values
band_2_peak = get_band_peak(band_2)
band_1_peak = get_band_peak(band_1)

File: src/find_ice_labels.jl [hunk 3]
```diff
@@ -165,6 +175,12 @@ function IceDetectionLopezAcosta2019(;
     ])
 end
 
+
+
+
+
+
+
```
cpaniaguam, line 183: Why all these blank lines?

cpaniaguam (commented):
This is great, thanks! Left a couple observations.

src/find_ice_labels.jl
danielmwatkins: very cool, good to know!

src/find_ice_labels.jl
danielmwatkins: That is cleaner, yeah!

src/find_ice_labels.jl
danielmwatkins: Ha, no reason. I'll clean that out.

src/histogram_equalization.jl
```diff
@@ -323,47 +323,4 @@ Histogram equalization of `img` using `nbins` bins.
...
-
-"""
-function imhist(img, imgtype::AbstractString="uint8")
-
-    # TODO: add validation for arr: either uint8 0:255 or grayscale 0:1
-    rng = imgtype == "uint8" ? range(0, 255) : range(0; stop=1, length=256)
-
-    return _imhist(img, rng)
-end
+end
```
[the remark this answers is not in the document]
danielmwatkins: Are you supposed to include an empty whitespace at the end of a file?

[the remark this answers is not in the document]
cpaniaguam: Not a white space but an empty line. It's kind of silly but github expects it. It's a rule in the POSIX standard. Most file formatters add it.

I think one good reason for observing this rule is that one might want to programmatically concatenate text files vertically. If there is no new line at the end, the last line of the first file will get fused with the first line of the second file.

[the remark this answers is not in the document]
danielmwatkins: Interesting, thanks for the explanation!

src/find_ice_labels.jl
```diff
@@ -202,4 +218,14 @@ end
...
\ No newline at end of file
+end
+
+function tiled_adaptive_binarization(img, tiles)
+    canvas = zeros(size(img))
+    for tile in tiles
+        f = AdaptiveThreshold(img[tile...])
+        canvas[tile...] = binarize(img[tile...], f)
+    end
+    return canvas
```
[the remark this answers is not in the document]
danielmwatkins: There are some steps later that are picky about black and white vs Bool vs BitMatrix. Here, I think letting the binarize function dictate the type is fine.

src/ice_masks.jl
```diff
@@ -187,11 +116,25 @@ function get_ice_masks(
...
+                band_7_max=thresholds[1],
+                band_2_min=thresholds[2],
+                band_1_min=thresholds[3]
+            ),
+        ) .|>
+        gray .|>
+        Bool
+    @debug "Found $(sum(mask)) ice pixels"
+    return mask
+end
```
hollandjg, line 140: I'd suggest including a newline at the end of every file, just for consistency.
```suggestion
end

```

src/histogram_equalization.jl
hollandjg, line 326:
```suggestion
end

```

hollandjg (approved):
This is great!
I've just a couple of comments around the edge. Carlos covered the important stuff.

It might be nice to make the tiled binarization into its own Meta-Binarization algorithm (which itself has a binarization), but that doesn't need to happen here. It might be too meta, too.

test/test-segmentation-lopez-acosta-2019-tiling.jl
```diff
@@ -90,11 +90,11 @@
         @test results_invariant_for(RGB; baseline, algorithm, case)
         @test results_invariant_for(RGBA; baseline, algorithm, case)
         @test results_invariant_for(n0f8; baseline, algorithm, case)
-        @test results_invariant_for(n6f10; baseline, algorithm, case) broken = true
-        @test results_invariant_for(n4f12; baseline, algorithm, case) broken = true
-        @test results_invariant_for(n2f14; baseline, algorithm, case) broken = true
-        @test results_invariant_for(n0f16; baseline, algorithm, case) broken = true
-        @test results_invariant_for(float32; baseline, algorithm, case) broken = true
-        @test results_invariant_for(float64; baseline, algorithm, case) broken = true
```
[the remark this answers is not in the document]
danielmwatkins: Yeah, it's nice to see that the update made the performance more stable!

PR #712, opened 2025-09-02
fix author email for Timothy Divoll in Project.toml

cpaniaguam:


hollandjg (approved):
Whoops. I missed that when I rearranged things!

PR #714, opened 2025-09-05
Refactor the Perona-Malik nonlinear diffusion function

danielmwatkins:
This pull requests introduces a nonlinear diffusion functor and reimplements the Perona-Malik diffusion algorithm based on a review of the original paper. It thus replaces the section of code taken from the Ultrasound Design Gallery. The function `IceFloeTracker.diffusion()` is replaced with `nonlinear_diffusion`, reflecting the view I've seen in the literature that Perona-Malik diffusion isn't truly anisotropic. The nested for loops are replaced with array operations. Like the form of the algorithm in the P-M paper, and like the Ultrasound Design Gallery function we were using, it uses the north-south-east-west, 4-nearest-neighbor differencing method for approximating the gradient.

Carlos's implementation of the Matlab anisotropic diffusion was moved from the `histogram_equalization` file into the new file `nonlinear_diffusion.jl`. His function was not being used in the code, but I've retained it in case we can use it to set up more options for the nonlinear_diffusion functor. It differs from the other approach in two ways: first, it uses the exponential form for the adaptive conductive flux term rather than an inverse quadratic function; second, it uses 8-nearest-neighbors method for the gradient. I've added an option to the Perona Malik diffusion functor so that either the exponential form or the inverse quadratic function can be used. A future update could set up the solver to use 8-nearest-neighbors instead.

Some questions:
* Is there a better way to handle options for the function `g`?
* Can we add an option to iterate until stability rather than iterating a set number of times?
* Can we estimate $K$ directly rather than prescribing it?


src/nonlinear_diffusion.jl
```diff
@@ -0,0 +1,185 @@
...
+    niters::Int = 3
+    g::String = "inverse_quadratic"
+    
+    # enforce conditions
+    function PeronaMalikDiffusion(λ, K, niters, g
+    )
+        !(0 < λ <= 0.25) && throw(ArgumentError("Lambda must be between zero and 0.25"))
+        K <= 0 && throw(ArgumentError("K must be greater than zero"))
+        niters <= 0 && throw(ArgumentError("Number of iterations must be greater than zero"))
+        g ∉ ["exponential", "inverse_quadratic"] && throw(ArgumentError("Unknown function name"))
```
cpaniaguam, line 36: I think this could be consolidated along with the actual functions into a single object. I'll add a PR with some suggestions.

src/nonlinear_diffusion.jl
```diff
@@ -0,0 +1,185 @@
...
+
+        # Back to regular 1-indexed arrays
+        flux_north_diff = flux_north[1:(end - 1), :] .- flux_north[2:end, :]
+        flux_east_diff = flux_east[:, 2:end] .- flux_east[:, 1:(end - 1)]
+
+        # Discrete PDE solution
+        sum_ = (1 / (dd^2)) .* (flux_nw .+ flux_ne .+ flux_sw .+ flux_se)
+
+        # Carlos - is this a typo? Shouldn't it be north minus east here? Or east minus north?
+        I = I .+ diffusion_rate .* (flux_north_diff .- flux_north_diff .+ sum_)
```
cpaniaguam, line 181: :scream:
I double-checked the MATLAB `\MATLAB\R2022a\toolbox\images\images\private\anisotropicDiffusion2D.m` file; it's actually the sum, north plus east!
```suggestion
        I = @. I + diffusion_rate * (flux_north_diff + flux_east_diff + sum_)
```
Thanks for catching that!

src/nonlinear_diffusion.jl
```diff
@@ -0,0 +1,185 @@
...
+    end
+
+    return I
+end
```
cpaniaguam, line 185:
```suggestion
end

```

cpaniaguam (commented):
Thanks for this and for catching the typo in the `anisotropic_diffusion_2D` function! I will add a PR with a couple suggestions that are a bit too complex to include in a review.

hollandjg (commented):
This looks really nice! I've made a couple of comments, but nothing is critical at all. Thanks for the reorganization!

hollandjg:
The code coverage issue is a pain – I think it's just because you've moved some otherwise untested code (though I'm surprised that that particular code isn't flexed in the tests, so it's worth looking more into why.)

hollandjg (approved):
This looks good to me!

PR #722, opened 2025-09-08
Fix unsharp mask function

danielmwatkins:
The unsharp mask function included an undocumented requirement that the input image be an integer matrix: at the return stage, the image is rounded to Int type. This resulted in unwanted binarization when taking normed images as inputs.

I updated the code to only use one clamp function and to recast the image to the input image type. I also added an option to use a threshold for applying the sharpening, so only regions where the difference between the image and the blurred image are large enough are changed.


danielmwatkins:
I could revert some things, but let's wait until the nonlinear diffusion functions are merged into main and then see how the tests turn out.

PR #723, opened 2025-09-08
ci[test]: remove lcov from codecov-action settings

hollandjg:
This is no longer specified in the example: https://github.com/codecov/example-julia/blob/main/.github/workflows/ci.yml 
No longer needed?


hollandjg:
> I'm not sure what this helps with beyond our having a license file in the main directory, so I don't mind taking things out.

As far as I can tell, this line is just no longer needed. This change just makes reading that file a 56th quicker.

PR #737, opened 2025-09-15
docs: add support for submodules

### LINKED ISSUE #733: Organize documentation of functions into sections
... like:
- morphology
- image filtering
- segmentation algorithms
- ...

Find out whether this is possible in an easy way (maybe modules?); if so include in release 0.10, if not move later.

hollandjg:
Add empty submodules which can be used for function reorganization, and update the documentation page to split the function listing.

The `@autodocs` tags can be included anywhere in the documentation as needed, so they don't have to be limited to the single page. 

- resolves #733


hollandjg:
> This looks great! Do you think we should get the functions added to modules before or after merging this pull request?

* I think it'll be easiest to move the functions step by step in a series of pull requests, probably making sure that the documentation and tests for each function work fine. That'll make each pull request that bit easier to parse. 
* Given that we're not releasing until we're done with this move, I think we can merge each pull request into main as we go. 
* (I tried starting with one of the larger functions, but there were so many moving pieces that I threw the branch away and started again here. I'd suggest doing the move through pull requests with size of the order of #738.)

PR #738, opened 2025-09-15
refactor: move @persist to utils submodule

hollandjg:
First example of reorganizing code in the IceFloeTracker – move the `@persist` macro to the Utils submodule.

- Removed dummy function in Utils submodule.
- Moved @persist and its dependencies into the module
- Export only the functions we use externally – in this case just the `@persist` macro.
- Re-export the `@persist` macro from IceFloeTracker.jl to support the existing uses. 
  (We could also require people to do `using IceFloeTracker.Utils: @persist` to load the macro. I prefer re-exporting, because it means that we get the internal organization, but we don't need to update all of the example and testing code.)


danielmwatkins (approved):
Good place to start!

PR #749, opened 2025-09-17
refactor: 688 unify `_process_image_tiles` and `conditional_histeq`

### LINKED ISSUE #688: Unify `_process_image_tiles` and `conditional_histeq`
I'm not sure the motivation for having the near-identical functions of `_process_image_tiles` and the `conditional_histeq`. It is what it is for now, but let's look at that when we refactor this section.

_Originally posted by @danielmwatkins in https://github.com/WilhelmusLab/IceFloeTracker.jl/pull/685#discussion_r2274496305_

hollandjg:
Make a single `conditional_histeq` function which doesn't directly compute tiles, but takes a `tiles` argument. 

Based on the uses we have, (and there aren't many) it seems sensible to leave the tile calculations outside of this function. We may want to add other versions of the function which calculates tiles, and this way we allow the user to use whichever version they want and pass in the tiles. 

- resolves #688

### DIFF
--- a/src/find_ice_labels.jl
+++ b/src/find_ice_labels.jl
@@ -217,12 +217,32 @@ function get_ice_labels(ice::AbstractArray{<:AbstractGray})
     return findall(vec(gray.(ice)) .> 0)
 end
 
-function tiled_adaptive_binarization(img, tiles)
+# TODO: Find the right home for this function
+# TODO: Set up wrapper for applying binarization to a tiled iterator. Let this
+# method be one functor algorithm option, and the version that uses the Peaks as a second.
+"""
+tiled_adaptive_binarization(img, tiles; minimum_window_size=). 
+
+    Applies the (AdaptiveThreshold)[https://juliaimages.org/ImageBinarization.jl/v0.1/#Adaptive-Threshold-1] binarization algorithm
+    to each tile in the image. Following the recommendations from ImageBinarization, the default is to use the integer window size
+    nearest to 1/8th the tile size if the tile is large enough. So that the window is large enough to include moderately large floes,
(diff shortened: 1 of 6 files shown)


src/histogram_equalization.jl
```diff
@@ -77,28 +77,54 @@ Convert an array of RGB channel data to grayscale in the range [0, 255].
...
+    entropy_threshold::Real=4.0,
+    white_threshold::Real=25.5,
+    white_fraction_threshold::Real=0.4,
 )
 
     # Apply Perona-Malik diffusion to each channel of true color image 
     # using the default inverse quadratic flux coefficient function
-    true_color_diffused = IceFloeTracker.nonlinear_diffusion(float64.(true_color_image), 0.1, 75, 3)
+    true_color_diffused = IceFloeTracker.nonlinear_diffusion(
+        float64.(true_color_image), 0.1, 75, 3
```
danielmwatkins, line 126: I ran the optimization against the matlab diffusion image, and the best parameters are 0.1, 0.1, and 5. Matlab also defaults to exponential, not quadratic. You can import the PeronaMalikDiffusion functor and then replace this line with
```julia
true_color_diffused = IceFloeTracker.nonlinear_diffusion(
        float64.(true_color_image), PeronaMalikDiffusion(0.1, 0.1, 5, "exponential"))
```

danielmwatkins (approved):
This looks good to me! 

As a future note, I wonder if this could be replaced with setting up a function to filter the tile iterator, apply the hist eq to the result, then merge the adjusted and original images. In that way, the choice to use the `clouds_red` object, the entropy thresholds, and all that are part of the filter function rather than the histogram adjustment function.

Issue #662, opened 2025-07-30
Reorganize functions by type


danielmwatkins:
Functions are not always in the place one might expect them to be. This is a checklist of functions that I subjectively think should be placed somewhere else.

- [x] `segmentation_a_direct.jl/kmeans_segmentation` ==> Julia file with segmentation algorithms (new API)
- [x] `segmentation_a_direct.jl/get_holes` ==> `fill_holes.jl` (or even a `morphology.jl` file)
- [x] `segmentation_a_direct.jl/fillholes!` ==> `fill_holes.jl`
- [x] `segmentation_a_direct.jl/branchbridge` ==> Morphology section
- [x] `segmentation_a_direct.jl/imgradientmag` ==> Image filtering section #804 
- [x] `histogram_equalization.jl/anisotropic_diffusion_2D/3D` ==> `nonlinear_diffusion.jl`
- [ ] `histogram_equalization.jl/get_rgb_channels` ==> `utils.jl` (if needed at all at the end of the day)
- [ ] `histogram_equalization.jl/rgb2gray` ==> Identical method to `Gray()` with the exception that it is all in UINT8.
- [ ] `normalization.jl/normalize_image` ==> not a normalization function, rather, it's morphology
- [ ] `normalization.jl/_adjust_histogram` ==> `histogram_equalization`
- [ ] `normalization.jl/imsharpen` ==> This is a preprocessing algorithm, with multiple steps. Perhaps a new file called `algorithms/preprocessing.jl` that has the preprocessing algorithms in it?
- [x] `normalization.jl/unsharp_mask` ==> this would fit better in an image filtering or contrast adjustment section
- [ ] `normalization.jl/imsharpen_gray` ==> we can delete this function #803 
- [ ] `normalization.jl/adjust_gamma` ==> This is a wrapper for the ImageContrastAdjustment adjust_histogram function. Just use `adjust_histogram` directly.
- [x] `normalization.jl/imbinarize` ==> Should be in a file with image binarization algorithms in it. This is just a call to the ImageBinarization `binarize` function.
- [ ] `imcomplement.jl` ==> Replace with calls to `complement()` from JuliaImages.
- [ ] `imadjust.jl` ==> Another wrapper of an ImageContrastAdjustment function, so  it should be in the `histogram_equalization.jl` file.
- [x] `ice_masks.jl` ==> This should be in a file with SegmentationAlgorithms in the new API
[...]

PR #750, opened 2025-09-17
refactor: move `imshow` to Utils.jl

hollandjg:
Move the `imshow` function to Utils, and simplify it.

In all the locations we use `imshow` in the examples, it is either used on a `BitMatrix`, _or_ the inputs are explicitly scaled up by a factor 255. We can simplify the code by just assuming that the inputs are within the range [0, 1] and not breaking that assumption.

part of #662


src/Utils/Utils.jl
```diff
@@ -103,6 +103,14 @@ function fname_ext_splice(fname::String, ext::String)
...
 end
 
-export @persist
+"""
+    imshow(array)
+
+Convert an array of booleans or reals in the range [0, 1] to a grayscale image.
+Values larger than 1 will be white, smaller than 0 will be black.
+"""
+imshow(array::Union{AbstractArray{<:Real,2},BitMatrix}) = Gray.(array) # function to display an array as a grayscale image
```
cpaniaguam, line 112: If the array is Int8 this won't work.
```julia
julia> img = rand(0:255, 10,10);

julia> typeof(img) <: Union{AbstractArray{<:Real,2}, BitMatrix}
true

julia> Gray.(img)
ERROR: ArgumentError: 254 is an integer in the range 0-255, but integer inputs are encoded with the N0f8
  type, an 8-bit type representing 256 discrete values between 0 and 1.
  Consider dividing your input values by 255, for example: Gray{N0f8}(254/255)
  Or use `reinterpret(N0f8, x)` if `x` is a `UInt8`.
  See the READMEs for FixedPointNumbers and ColorTypes for more information.
```

cpaniaguam (changes requested):
Just a couple of observations, thanks!

danielmwatkins:
Perhaps a naive question: what does `imshow` offer that `Images.Gray.()` doesn't already provide?

hollandjg:
> Perhaps a naive question: what does `imshow` offer that `Images.Gray.()` doesn't already provide?

From my point of view, assuming that people are already in the "Images.jl world" and know that `Gray.(foo)` will attempt to convert `foo` into an image, there's one benefit – you can apply `imshow` to a vector of arrays like this: 
```julia
imshow.(vector_of_arrays)
``` 
and get a row of images back. We use that in some of the example notebooks.

The alternative in those cases is to do 
```julia
[Gray.(img) for img in imgs]
```

...or define the `imshow` helper function in each of the notebooks independently. It's not a complicated function.
```julia
imshow(x) = Gray.(x)
```

`imshow` is perhaps not a function we need in the package. I'd be happy to remove it completely.

PR #751, opened 2025-09-17
refactor: move `imregionalmin`, `imextendedmin`, `bwdist` to morphology

hollandjg:
Start moving morphology functions.

- part of #662


src/Morphology/Morphology.jl
```diff
@@ -1,18 +1,48 @@
...
+- `img`: image object
+- `h`: suppress minima below this depth threshold
+- `conn`: neighborhood connectivity; in 2D 1 = 4-neighborhood and 2 = 8-neighborhood
 """
-function dummy_morphology_function()
-    return "This is a dummy morphology function."
+function imextendedmin(img::AbstractArray, h::Int=2, conn::Int=2)::BitMatrix
+    mask = hmin_transform(img, h)
+    mask_minima = local_minima(mask; connectivity=conn)
+    return mask_minima .> 0
```
cpaniaguam, line 36: These two lines can be replaced with a call to `imregionalmin`.

```suggestion
    mask_minima = imregionalmin(mask; connectivity=conn)
    return mask_minima
```

cpaniaguam (changes requested):
Thanks for this! Just one thing here.

src/Morphology/Morphology.jl
danielmwatkins: As a counterpoint, I think it would be more straightforward to keep using the ImageMorphology function `local_minima` rather than introduce a wrapper that essentially only renames the function to match the Matlab function. If we were extending the `local_minima` to have deeper functionality then I think it makes sense to have another function. If we really need only one line you could replace

```julia
mask_minima = local_minima(mask; connectivity=conn)
return mask_minima .> 0
```

with

```julia
return local_minima(mask; connectivity=conn) .> 0
```
(or use convert(BitMatrix, ) if you want).

danielmwatkins:
Regarding `bwdist` -- this is used to set things up for segmentation via watershed, and is I think nice to have as a separate function vs the construction in Julia `ImageSegmentation`

```julia
dist = 1 .- distance_transform(feature_transform(bw));
```

It's not a morphological operation though. It's main use is for segmentation. So I think it should go in Segmentation.jl.

hollandjg:
Thanks for the feedback!
- [ ] move `bwdist` to `Segmentation.jl` instead
- [ ] leave suggested `imregionalmin` change out

hollandjg:
replaced by #765 for imregionalmin, imextendedmin

PR #765, opened 2025-10-01
gather morphology functions in Morphology.jl submodule

hollandjg:


danielmwatkins (approved):
Great first step, this is already helpful with making it easier to find particular functions. I see a few things in here that we can prune (e.g., I don't see the point of having the `imregionalmin` function in here instead of just using `local_minima` everywhere), but I think it's the right move to focus on organization alone with this pull request and pruning in a separate PR.

PR #769, opened 2025-10-14
fix precompiler errors in Julia 1.12

hollandjg:
The errors we've been having in CI recently may be to do with the release of Julia 1.12. This PR reverts the broken workflows to Julia 1.11.7 and can be reverted once 1.12 is stable.


danielmwatkins (approved):
Seems like a reasonable approach for now! Hopefully Julia 1.12 gets patched in the meantime.

Issue #766, opened 2025-10-07
Logic error in get_time_space_proximity_condition


danielmwatkins:
The 3-step proximity condition function will checks whether particular pairs of time steps and distances are allowed. The unexpected "gotcha" is that 

Here's the function:
```julia
function get_time_space_proximity_condition(
    d, delta_time, search_thresholds=(dt=(30, 100, 1300), dist=(15, 30, 120))
)
    return (delta_time < search_thresholds.dt[1] && d < search_thresholds.dist[1]) ||
           (
               delta_time >= search_thresholds.dt[1] &&
               delta_time <= search_thresholds.dt[2] &&
               d < search_thresholds.dist[2]
           ) ||
           (delta_time >= search_thresholds.dt[3] && d < search_thresholds.dist[3])
end
```

Notice that if floes are separated in time by 101 minutes and have traveled 0 pixels, with the default settings, we will get
1. delta_time < search_thresholds.dt[1] --> False
2. delta_time <= search_thresholds.dt[2]  --> False
3. delta_time >= search_thresholds.dt[3] --> False

But if the floes were 1300 minutes apart and had the same travel distance, condition 3 would have evaluated as true.

Issue #588, opened 2025-03-10
decide whether get_time_space_proximity_condition should be private


cpaniaguam:
```
# TODO: decide whether get_time_space_proximity_condition should be private
```

In file src/tracker/tracker-funcs.jl
https://github.com/WilhelmusLab/IceFloeTracker.jl/blob/34a28427ae558f210324e0fcde90ee18ac6d4815/src/tracker/tracker-funcs.jl#L185

danielmwatkins:
I think the question here is whether we want to be able to change the way that the proximity is decided. The current version is pretty rigid, with it's three-tier approach. I'd like to have a continuous function instead, that is based on observational data.

danielmwatkins:
As with the `create_cloudmask()` function, this function can be upgraded to include a few different options for checking time and distance criteria.

PR #770, opened 2025-10-14
Introduce time-distance threshold functors

danielmwatkins:
It addresses the bug identified in #766 and resolves the question about proximity tests in #588. It also begins the task of moving tracking functionality into structural groups in the Tracking submodule. It adds a test for the time distance functors. And I update and clarify the threshold settings by renaming `condition_thresholds` and `mc_thresholds` to `candidate_filter_settings` and `candidate_matching_settings`. These new named tuples generalize the idea of thresholds to enable us to pass functions into the tracker. Down the road, I'd like to generalize the small floe / large floe dichotomy to allow area-dependent thresholds.

### DIFF
--- a/src/tracker/tracker-funcs.jl
+++ b/src/tracker/tracker-funcs.jl
@@ -1,7 +1,7 @@
 # need this for adding methods to Base functions
 import Base.isempty
 import Base.isequal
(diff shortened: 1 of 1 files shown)


hollandjg (approved):
Looks good to me!

PR #779, opened 2025-10-16
Reorganize tracker functions

danielmwatkins:
This pull request moves the main tracker functions into the Tracking module. For the most part, this involves simply moving files into the folder and adding exports to Tracking.jl. For region_props, I've taken the step to take the portion of the code in regionprops.jl that is only used in tracking -- cropfloes and related utilities -- and added that to a new file extend_regionprops. In the next step, I'll be moving functions from tracker-funcs into extend_regionprops, with the idea that we use regionprops to summarize segmented image output, and extend_regionprops to add psi-s and add floe masks. 

As a note for later: I think that the passtimes and latlon information should be added to the regionprops table in the Segmentation module. The latlon and polar stereographic X/Y data is critical for making sure that the geometric measurements have the proper units, and the time information can be inherited from the image in most cases.

### DIFF
--- a/src/Tracking/psi-s.jl
+++ b/src/Tracking/psi-s.jl
@@ -121,7 +121,7 @@ function make_psi_s(
         phase = atan.(dy, dx)
     end
 
-    if unwrap
+    if dsp_unwrap
         phase = unwrap(phase)
     end
 
@@ -140,5 +140,5 @@ Alternate method of `make_psi_s` accepting input vectors `x` and `y` as a 2-colu
 function make_psi_s(XY::Matrix{<:Number}; rangeout::Bool=true, dsp_unwrap::Bool=true)
     x = XY[:, 1]
     y = XY[:, 2]
-    return make_psi_s(x, y; rangeout=rangeout, unwrap=dsp_unwrap)
+    return make_psi_s(x, y; rangeout=rangeout, dsp_unwrap=dsp_unwrap)
 end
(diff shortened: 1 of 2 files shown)


danielmwatkins:
This is ready to go! I don't know why the codecov is failing, all the functions in regionprops.jl have docstrings. We'll be revisiting that script anyway when we move it into Segmentation.jl so I think we should merge this anyway.

src/Tracking/crosscorr.jl
```diff
@@ -91,6 +92,8 @@ julia> [r lags]
 function crosscorr(
     u::Vector{T}, v::Vector{T}; normalize::Bool=false, padmode::Symbol=:longest
 )::Tuple{Vector{T},Vector{T}} where {T<:Real}
+    # dmw: check whether we could implement this with the tools in StatsBase or Statistics
+    # so we can avoid an extra import.
     c = DSP.xcorr(u, v; padmode=padmode)
```
hollandjg:
```suggestion
    c = xcorr(u, v; padmode=padmode)
```

src/Tracking/register.jl
```diff
@@ -1,3 +1,5 @@
+using Images: imrotate, padarray, Fill
```
hollandjg:
```suggestion
import Images: imrotate, padarray, Fill
```

src/Tracking/crosscorr.jl
```diff
@@ -1,3 +1,4 @@
+using DSP
```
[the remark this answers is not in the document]
danielmwatkins: I like that approach. I've been using `using` in imitation of the other functions in the package, but I agree that it would be better to use import and be strict about what we are pulling into the namespace. I've noticed there are a few conflicts already -- not causing problems but raising warnings.

PR #780, opened 2025-10-17
Adding segmentation tracking notebook to tutorials folder

pheladurosinmi:
Pull request to add the combined segmentation and tracking notebook.


segmentation-tracking-combined.ipynb
danielmwatkins: Check your folder for this file, it's empty but was included in the commit.

pheladurosinmi:
<p>Currently I am unable to combine all package imports at once when running the code as it causes my kernel to die, however, when I push the next version I will combine all.</p>

---

 <a href='https://app.reviewnb.com/WilhelmusLab/IceFloeTracker.jl/pull/780/discussion/#comment-3427057696'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

danielmwatkins:
Good progress! I'm going to put my notes here rather than inside the ReviewNB app, so that you don't get 30 different notifications again.

I like the inclusion of a link at the beginning and the guideposts for the notebook.

For the "Environment Setup" section, let's use just one block for the package installation and one for the "using" calls rather than a bunch of blocks. 

```julia
using Pkg
Pkg.activate("ift_demo")
Pkg.add(;name="IceFloeTracker", rev="main")
Pkg.add("Images")
Pkg.add("Plots")
```

When the next release happens, we can take out the "rev=main" part. Explicitly adding an environment (the "activate") call above can help with managing conflicts between packages.

```julia
using IceFloeTracker
using Images
using Plots
```

Here is a nice way to load the data:
```julia

data_loader = Watkins2025GitHub(; ref="b865acc62f223d6ff14a073a297d682c4c034e5d")

cn = 6
dataset = data_loader(c-> c.case_number == cn)
cases = Dict((case.name => case) for case in dataset)
```

In this way, the dictionary has an entry for each case name.

With this structure, we can plot the cases
```julia
mosaicview([cases[c].modis_truecolor for c in keys(cases)], nrow=1, npad=10)
```

When you call the segmentation algorithm, you can do the same kind of "list comprehension":
```julia
segments = [segmentation_algorithm(
                RGB.(cases[c].modis_truecolor),
                RGB.(cases[c].modis_falsecolor),
                RGB.(cases[c].modis_landmask)) for c in keys(cases)];
```

and you can update the mosaicview part to have indices
```
mosaicview(view_seg(segments[1]), view_seg(segments[2]), nrow=1, npad=10)
```

I don't think you need to show the binary images, it just adds a complication. 

In the Tracking Algorithm section, you can note that if one is starting with binary images, you can use `label_components` to assign a unique label to each region. Here, though, we can use the `image_indexmap` from the segmented images.

```julia
props = regionprops_table.([segments[1].image_indexmap, segments[2].image_indexmap])
```

You'll need to update the passtimes vector from the case metadata, e.g. `cases[c].metadata[:pass_time]`.

(I'll have to update the names at some point since the tracker wants it to be called :passtime).

We can talk over the plotting section in person. For now -- remember when you are making a pull request that you don't want there to be any errors remaining when you commit it. So either resolve errors or remove sections that cause errors.

PR #782, opened 2025-10-22
refactor: move Segmentation functions to submodule

hollandjg:
- Move functions to the Segmentation submodule.
- Move landmasking to Preprocessing
- Split the k-means segmentation from its current file and include in Segmentation
- Make an ImageUtils submodule including just the `masker`
- Make a Geospatial submodule with `latlon`


danielmwatkins (changes requested):
I think this is a good start for this section of the reorganization. I like the idea of the geospatial module since that has a lot of specific imports not used elsewhere, and I wonder what other things we might want to use there. Splitting Utils and ImageUtils makes sense. With the landmask function, I wonder if that should go into preprocessing or into geospatial. I don't think it's quite right to go in Morphology -- it uses morphological operations, but "making a landmask" isn't itself a morphological operation, whereas most of the other functions there are more use-case agnostic. If in the future we want to be able to import landmask shape files and translate them into binary images, then that functionality would make more sense in Geospatial.

danielmwatkins (approved):
Looks good to me!

PR #785, opened 2025-10-22
deps: remove Pkg dependency

hollandjg:
Remove unused `Pkg` import in package.


danielmwatkins (approved):
Interesting, was there a point where we were actively installing new packages as part of the package?

PR #786, opened 2025-10-22
deps: remove Serialization dependency

hollandjg:
Move it to be a dependency of the Notebook and the Tests – it's not used in the actual package so we don't need to import it.


danielmwatkins (approved):
I find the deserialize/serialize workflow opaque anyway, so I tend to avoid it (as I also avoid Python’s pickle function). I think for the cases we’re demonstrating in the notebooks, the next step is to use labeled images and `extend_regionprops` to put the materials together for tracking.

PR #787, opened 2025-10-22
deps: remove DelimitedFiles dependency

hollandjg:
The test environment requires delimited files, but the main package doesn't.


danielmwatkins (commented):
Is there a reason that we need DelimitedFiles and CSV and CSVFiles? Would it be enough to use FileIO and CSVFiles throughout?

hollandjg:
> Is there a reason that we need DelimitedFiles and CSV and CSVFiles? Would it be enough to use FileIO and CSVFiles throughout?

I think just using FileIO and CSVFiles would be great. I've opened an issue.

PR #808, opened 2025-10-31
fix: PyCall installation

hollandjg:
Remove extra code in PyCall setup. The bugfix from before seems to be broken.


danielmwatkins:
Weird how PyCall will suddenly break sometimes. On my pull request earlier today, it was working all morning then stopped -- but only for two of the CIs.

cpaniaguam (approved):
Lgtm

PR #810, opened 2025-11-03
refactor: use PythonCall rather than PyCall

hollandjg:
This PR replaces PyCall with PythonCall. 
- Alternative to #808 

Benefits:
- PythonCall makes an isolated Python environment for each project the end-user has, which means that we reduce the chance of incompatibility with python package versions the user needs for other projects.
- PythonCall doesn't need extra setup steps, and when given the new CondaPkg.toml file added in this PR, seems to work out of the box. (So we can remove extra installation steps in the project README and remove those steps from the GitHub workflows.)
- PythonCall is maintained (last updated in September '25) whereas PyCall wasn't (last updated in December '23).
- Seems to work with Julia 1.12.1.

Drawbacks:
- PythonCall and PyCall aren't immediately compatible with each other in the same project – you have to set the conda environment explicitly.
- It embeds PythonCall more thoroughly in the package than PyCall was, because we have to do explicit conversions between Julia and Python types. (PyCall does this automatically.) 
  > One way we could counteract that is to isolate the PythonCall parts in the `skimage` module, where we'd make "wrapped" versions of the functions we're using which use Julia alone on the interfaces. We could do that now or later.

Additional validations:
- This works with Colab notebooks. 
  (Open [the Colab notebook](https://colab.research.google.com/github/WilhelmusLab/IceFloeTracker.jl/blob/refs/heads/main/docs/src/tutorials/lopez-acosta-2019-workflow.ipynb) and insert `Pkg.add(;name="IceFloeTracker",rev="refactor--use-PythonCall")` in the first cell to test. Precompiling took ~2000s.)

My hope with switching from PyCall to PythonCall is that we'll have fewer, rarer problems with Python in future.

### DIFF
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -1,3 +1,3 @@
 [deps]
-python = "=3.12"
+python = ">=3.12,<4"
 scikit-image = "=0.25.2"


src/Pipeline/LopezAcosta2019Tiling.jl
```diff
@@ -490,8 +492,12 @@ function get_final(
 
     #= opening to remove noise while preserving shape/size
     Note the different structuring elements for erosion and dilation =#
-    mask = sk_morphology.erosion(_img, se_erosion)
-    mask .= sk_morphology.dilation(mask, se_dilation)
+    mask = pyconvert(
+        Array, sk_morphology.erosion(Py(_img).to_numpy(), Py(se_erosion).to_numpy())
+    )
+    mask .= pyconvert(
+        Array, sk_morphology.dilation(Py(mask).to_numpy(), Py(se_dilation).to_numpy())
```
danielmwatkins: Any ideas why we would need the sk_morphology erosion and dilation functions? One possibility would be if the sk_morphology functions are faster for unusual structuring elements.

tdivoll (commented):
I don't have too much to add other than I agree it makes sense to keep this PR focused on the translation from PyCall to PythonCall and make new PRs for the other features. If I remember, we had trouble getting IFT to use the right conda env unless we used PyCall and let it use Julia's default conda env. It seems like PythonCall has better support now and the CondaPkg toml file makes it explicit?

PR #819, opened 2025-11-06
refactor: replace `skimage.morphology.{dilation,erosion}` with Images.jl `{dilate,erode}`

hollandjg:
Replacing sk_morphology.{dilation,erosion} with Images.jl {dilate,erode}.

Seems to reduce our speed a bit. Is it too much?


cpaniaguam (changes requested):
There remain a few calls to [skimage.morphology.reconstruction](https://scikit-image.org/docs/0.25.x/api/skimage.morphology.html#skimage.morphology.reconstruction) in the file, which can potentially be replaced with `ImageMorphology.mreconstruction` (with dilation).

cpaniaguam (approved):

PR #821, opened 2025-11-10
Update tracker function to use list of tests

danielmwatkins:
- [x] Linear time-distance function/functor
- [ ] Update the tests to include use of ValidationDataset, including accurate passtimes
    - [x] Tests of the extend_regionprops + threshold test functions
    - [ ] Tests of the main tracker function
- [x] Introduce functions that add test variables to the props dataframe. These functions make use of the `transform!` function from DataFrames to apply functions to each row and save the output to a new column.
    - [x] distance thresholds --> adds `Δx, Δt`
    - [x] relative error functions (formerly absolute difference ratio) --> generic for a scalar measure `m`, adds column `relative_error_$m`
    - [x] shape difference --> adds `shape_difference`, `scaled_shape_difference` (scaled = divide by area or perimeter)
    - [x] psi-s correlation --> adds `ρ` and the confidence interval for ρ.
- [x] Update pairfloes to accept a named tuple of functions to test. Tuple names are the columns used for the test functions.
- [x] Introduce functions that decide between conflicting pairings to get a unique matching
- [ ] Make an example with default functions based on the calibration paper
- [ ] Cleanup: remove duplicate methods from Tracking module
- [x] Add direct shape difference calculation using the floe orientations so only two rotations are needed to get the shape difference metric (rotate each floe by orientation, then both are aligned on y axis).

### DIFF
--- a/src/Tracking/filter_functions.jl
+++ b/src/Tracking/filter_functions.jl
@@ -198,7 +198,7 @@ applying 7 individual AbstractFloeFilterFunctions in sequence:
 Filters 2-7 use PiecewiseLinearThresholdFunctions for thresholds, while Filter 1 uses a LinearTimeDistanceFunction.
 The default values and settings are derived in Watkins et al. 2026.
 """
-FilterFunction = ChainedFilterFunction(
+FilterFunction() = ChainedFilterFunction(;
     filters = [
         DistanceThresholdFilter(),
         RelativeErrorThresholdFilter(variable=:area,
--- a/notebooks/track-floes/example-track-floes-rotation.ipynb
+++ b/notebooks/track-floes/example-track-floes-rotation.ipynb
@@ -22,8 +22,16 @@
     "# Setup environment\n",
     "using Pkg\n",
     "Pkg.add(\"IceFloeTracker\")\n",
-    "Pkg.add([\"CSVFiles\", \"DataFrames\", \"Dates\", \"Plots\"])\n",
-    "\n",
+    "Pkg.add([\"CSVFiles\", \"DataFrames\", \"Dates\", \"Plots\"])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4ba801f4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "using IceFloeTracker\n",
     "using DataFrames, Dates, Plots, CSVFiles\n",
     "using Images: load, label_components\n",
@@ -147,7 +155,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "add_floemasks!(props, [im .> 0 for im in images]) # Images need to be bitmatrices for the floemasks call.\n",
+    "add_floemasks!(props, labeled_images)\n",
     "add_ψs!(props)"
    ]
   },
@@ -206,7 +214,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "display(\"text/markdown\", @doc long_tracker)"
+    "display(\"text/markdown\", @doc floe_tracker)"
    ]
   },
   {
@@ -224,7 +232,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "cf = IceFloeTracker.ChainedFilterFunction()"
+    "cf = IceFloeTracker.FilterFunction()"
    ]
   },
   {
@@ -381,15 +389,15 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Julia 1.12.1",
+   "display_name": "Julia 1.11.7",
    "language": "julia",
-   "name": "julia-1.12"
+   "name": "julia-1.11"
   },
   "language_info": {
    "file_extension": ".jl",
    "mimetype": "application/julia",
    "name": "julia",
-   "version": "1.12.1"
+   "version": "1.11.7"
   }
  },
  "nbformat": 4,
--- a/notebooks/track-floes/track-floes.ipynb
+++ b/notebooks/track-floes/track-floes.ipynb
@@ -66,7 +66,7 @@
    ],
    "source": [
     "# Initialize functions using default settings\n",
-    "filter_funs = IceFloeTracker.ChainedFilterFunction()\n",
+    "filter_funs = IceFloeTracker.FilterFunction()\n",
     "matching_fun = IceFloeTracker.MinimumWeightMatchingFunction()"
    ]
   },
--- a/test/test-floe_tracker.jl
+++ b/test/test-floe_tracker.jl
@@ -109,7 +109,7 @@ end
         # Every floe is matched in every day
         props_test_case1 = deepcopy(_props)
         trajectories = floe_tracker(
-            props_test_case1, FilterFunction, MinimumWeightMatchingFunction()
+            props_test_case1, FilterFunction(), MinimumWeightMatchingFunction()
         )
 
         # Expected: 5 trajectories, all of which have length 3
@@ -126,7 +126,9 @@ end
 
     @testset "Case 2" begin
         trajectories = IceFloeTracker.floe_tracker(
-            props_test_case2, FilterFunction, IceFloeTracker.MinimumWeightMatchingFunction()
+            props_test_case2,
+            FilterFunction(),
+            IceFloeTracker.MinimumWeightMatchingFunction(),
         )
 
         # Expected: 5 trajectories, 3 of which have length 3 and 2 of which have length 2
@@ -142,7 +144,7 @@ end
             props = addgaps(_props)
 
             trajectories = IceFloeTracker.floe_tracker(
-                props, FilterFunction, MinimumWeightMatchingFunction()
+                props, FilterFunction(), MinimumWeightMatchingFunction()
             )
 
             # Expected: 5 trajectories, all of which have length 3 as in test case 1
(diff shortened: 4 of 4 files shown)


hollandjg:
> There's one test, `test-register-utils`, that looks like it might be sensitive to roundoff error - in line 616 it tested if the shape difference is exactly 33. Turns out it's 32 on my machine and 33 when we run the test online. Since the rotation involves interpolation, I suspect we'd actually want the answer to be ± 1.

Curious. Are the two machines Windows and macOS? 

Could you run the function on both machines and check where the difference is?:
```julia
using IceFloeTracker
img = Bool[
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 1 1 1 0 0 0 0 0 0 0 0
                0 0 0 0 1 1 1 1 1 0 0 0 0 0 0 0
                0 0 0 1 1 1 1 1 1 1 0 0 0 0 0 0
                0 0 0 0 1 1 1 1 1 1 1 0 0 0 0 0
                0 0 0 0 0 1 1 1 1 1 1 1 0 0 0 0
                0 0 0 0 0 0 1 1 1 1 1 1 1 0 0 0
                0 0 0 0 0 0 0 1 1 1 1 1 1 1 0 0
                0 0 0 0 0 0 0 0 1 1 1 1 1 0 0 0
                0 0 0 0 0 0 0 0 0 1 1 1 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
            ]
rotated_img = IceFloeTracker.Tracking.imrotate_bin_clockwise_radians(img, -Float64(π / 4.0))
```

I'm a bit surprised that the BSpline(Constant()) we're using produces platform-dependent differences at the edges, but perhaps I shouldn't be. It might be a reason to use a higher order spline and to go to fraction-valued masks rather than the binary ones we're using now.

src/Tracking/distance_thresholds.jl
```diff
@@ -1,10 +1,17 @@
 import Dates: Millisecond, Second, Minute, Hour, Day
-abstract type AbstractTimeDistanceThresholdFunction end
+abstract type AbstractThresholdFunction <: Function end
```
hollandjg: Aha! That's how you do it.

src/Tracking/extend_regionprops.jl
```diff
@@ -1,11 +1,90 @@
...
+    df.uuid = [_uuid() for _ in 1:nrow(df)]
+    return df
+end
+
+function add_uuids!(dfs::Vector{DataFrame})
+    for (i, _) in enumerate(dfs)
+        add_uuids!(dfs[i])
+    end
+    return dfs
+end
```
hollandjg: There's a clever way to avoid having to write out these vectorizing functions each time. Given:
```julia
dfs::Vector{DataFrame}
```
you can broadcast over all the elements in the vector with:
```julia
add_uuids!.(dfs)
```

src/Tracking/filter_functions.jl
```diff
@@ -0,0 +1,218 @@
...
+
+"""
+    LopezAcosta2019ChainedFilterFunction(floe, candidates)
+
+The LopezAcosta2019ChainedFilterFunction is a special case of ChainedFilterFunction
+with parameters and threshold functions set based on Lopez-Acosta et al. 2019. The set
+of threshold filters is the same as in the default ChainedFilterFunction, but using
+stepwise threshold functions instead of piecewise.
+"""
+LopezAcosta2019ChainedFilterFunction = ChainedFilterFunction(
```
hollandjg: I like this example function. We can move it to the `LopezAcosta2019` submodule later,

danielmwatkins:
Actually I thought of a different difference - I have Julia 1.12 on my home computer. Hmm. And yes I think going to fractional values is going to be an important improvement. I also implemented a shape difference function that rotates floes by their orientation first. I have to think through how the angles work to see if that can be a starting point for a faster search. It’s definitely faster for just directly getting the minimal shape difference though.-DanielOn Nov 18, 2025, at 16:33, John Gerrard Holland ***@***.***> wrote:﻿hollandjg left a comment (WilhelmusLab/IceFloeTracker.jl#821)

There's one test, test-register-utils, that looks like it might be sensitive to roundoff error - in line 616 it tested if the shape difference is exactly 33. Turns out it's 32 on my machine and 33 when we run the test online. Since the rotation involves interpolation, I suspect we'd actually want the answer to be ± 1.

Curious. Are the two machines Windows and macOS?
Could you run the function on both machines and check where the difference is?:
using IceFloeTracker
img = Bool[
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 1 1 1 0 0 0 0 0 0 0 0
                0 0 0 0 1 1 1 1 1 0 0 0 0 0 0 0
                0 0 0 1 1 1 1 1 1 1 0 0 0 0 0 0
                0 0 0 0 1 1 1 1 1 1 1 0 0 0 0 0
                0 0 0 0 0 1 1 1 1 1 1 1 0 0 0 0
                0 0 0 0 0 0 1 1 1 1 1 1 1 0 0 0
                0 0 0 0 0 0 0 1 1 1 1 1 1 1 0 0
                0 0 0 0 0 0 0 0 1 1 1 1 1 0 0 0
                0 0 0 0 0 0 0 0 0 1 1 1 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
                0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
            ]
rotated_img = IceFloeTracker.Tracking.imrotate_bin_clockwise_radians(img, -Float64(π / 4.0))
I'm a bit surprised that the BSpline(Constant()) we're using produces platform-dependent differences at the edges, but perhaps I shouldn't be. It might be a reason to use a higher order spline and to go to fraction-valued masks rather than the binary ones we're using now.

src/Tracking/filter_functions.jl
```diff
@@ -0,0 +1,218 @@
...
+    filters = [
+        DistanceThresholdFilter(),
+        RelativeErrorThresholdFilter(variable=:area),
+        RelativeErrorThresholdFilter(variable=:convex_area),
+        RelativeErrorThresholdFilter(variable=:major_axis_length),
+        RelativeErrorThresholdFilter(variable=:minor_axis_length),
+        ShapeDifferenceThresholdFilter(),
+        PsiSCorrelationThresholdFilter()
+        ]
+end
```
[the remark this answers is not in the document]
danielmwatkins: I've moved this into a new FilterFunction as suggested.

src/Tracking/extend_regionprops.jl
danielmwatkins: Yes, let's do that in the next PR

src/Tracking/register.jl
```diff
@@ -70,6 +70,46 @@ function align_centroids(im1::AbstractArray{Bool}, im2::AbstractArray{Bool})
...
+
+"""
+function shape_difference(floe1, floe2)
+    im1 = imrotate_bin_nocrop(floe1.mask, floe1.orientation)
+    im2 = imrotate_bin_nocrop(floe2.mask, floe2.orientation)
+    im1, im2 = align_centroids(im1, im2)
+    a_not_b = im1 .> 0 .&& isequal.(im2, 0)
+    b_not_a = im2 .> 0 .&& isequal.(im1, 0)
+    shape_difference = sum(a_not_b .|| b_not_a)
+    return shape_difference
```
hollandjg: Deduplicate code by using a call like this?
```suggestion
    shape_difference_ = shape_difference(floe1.mask, floe1.orientation, floe2.mask, floe2.orientation)
    return shape_difference_
```

src/Tracking/register.jl
```diff
@@ -70,6 +70,46 @@ function align_centroids(im1::AbstractArray{Bool}, im2::AbstractArray{Bool})
     return im1_padded, im2_padded
 end
 
+
+"""
+    shape_difference(floe1::DataFrameRow, floe2::DataFrameRow)
+    
+    Computes the shape difference between two ice floes using the estimated floe
+    orientation to first align both floes on the same axis. 
+
```
hollandjg:
```suggestion
    shape_difference(floe1::DataFrameRow, floe2::DataFrameRow)
    shape_difference(floe1_mask::BitMatrix, floe1_orientation::Float64,
                     floe2_mask::BitMatrix, floe2_orientation::Float64)
    
    Computes the shape difference between two ice floes using the estimated floe
    orientation to first align both floes on the same axis. 
    
    Passing the mask and orientation
    together lets us more efficiently use it for DataFrame transforms.

```

danielmwatkins:
OK cool! I'll let the tests finish on the PR and then get it merged here. It would be great to get this PR finished!

PR #830, opened 2025-11-24
refactor: rewrite Watkins2026Dataset to use accessor functions

hollandjg:
No need to instantiate a separate "loader", now we just do:
```julia
dataset = Watkins2026Dataset()
case = first(dataset)
img = modis_truecolor(case)
```

Filtering can be performed directly on any dataset, and both `filter` and `DataFrames.subset` are supported:
```julia
filtered_dataset = filter(c-> c.case_number==6, dataset)
# equivalent to:
subsetted_dataset = subset(dataset, :case_number => n -> n .== 6)
```

Images are accessed using methods (and not properties):
```julia
image::AbstractArray{RBGA, 2} = modis_truecolor(case)
```

... and vectors of images can be accessed using broadcasting notation:
```julia
images::Vector{AbstractArray{RBGA, 2}} = modis_truecolor.(dataset)
```

... so that multi-image segmentation is easy to perform:
```julia
segmenter = LopezAcosta2019.Segment()
segmented_images = segmenter.(modis_truecolor.(dataset), modis_falsecolor.(dataset), modis_landmask.(dataset))
```

### DIFF
--- a/notebooks/preprocessing-workflow/lopez-acosta-2019-workflow.ipynb
+++ b/notebooks/preprocessing-workflow/lopez-acosta-2019-workflow.ipynb
@@ -32,7 +32,7 @@
    "outputs": [],
    "source": [
     "# Load packages\n",
-    "using IceFloeTracker: LopezAcosta2019, LopezAcosta2019Tiling, Watkins2026Dataset, metadata, modis_truecolor, modis_landmask, modis_falsecolor, validated_binary_floes\n",
+    "using IceFloeTracker: LopezAcosta2019, LopezAcosta2019Tiling, Watkins2026Dataset, info, modis_truecolor, modis_landmask, modis_falsecolor, validated_binary_floes\n",
     "using Images: erode, segment_mean, labels_map, SegmentedImage, RGB, mosaicview"
    ]
   },


danielmwatkins:
Nice! In this new workflow, does it still do lazy loading where the images are downloaded when the particular case is selected? Or do they all get downloaded at first instantiation?

Using the methods to load the data does seem like a more Julia approach, which is cool.

hollandjg:
> Nice! In this new workflow, does it still do lazy loading where the images are downloaded when the particular case is selected? Or do they all get downloaded at first instantiation?

It's even better than that – the images are downloaded when you first try to use that particular image, so if you want just the truecolor MODIS image from a case, only that image will be downloaded. In the past, downloading a case meant downloading all of the components (truecolor, falscolor, masie, etc.). We still get the cacheing we had before.

danielmwatkins:
Also as a heads up, I published the first release for the ice floe validation dataset repo, so it should be copying over to Zenodo now. So we can add the setting to get the repo based on release number now.

hollandjg:
> Also as a heads up, I published the first release for the ice floe validation dataset repo, so it should be copying over to Zenodo now. So we can add the setting to get the repo based on release number now.

I've updated everywhere to use v0.1!

PR #835, opened 2025-11-26
Unify k-means binarization across the two Pipelines

danielmwatkins:
This pull request introduces a new (replacement) k-means segmentation method and a k-means binarization function. The function takes an IceDetectionAlgorithm as an input. It replaces the previous methods, where `kmeans_segmentation` included both the standard clustering result and a binarized result. This change makes it so that we can define IceDetectionAlgorithms in each of the Pipelines, and thus not need the separate `kmeans_segmentation` and `get_ice_masks/get_nlabel` workflows.

I've introduced a new function, `stitch_clusters`, which is aimed at addressing the issue where k-means clusters in the tiling result can have sharp discontinuities. I've implemented a test function for it. However the tiled kmeans binarization function doesn't use it yet - I wanted to get everything else running first.

This pull request is a step toward streamlining the pipelines -- I found that the PR focused on the LopezAcosta2019 algorithm kept having mission creep as functions were being used in both algorithms. So hopefully after this PR goes through it's simpler to refactor the non-tiling algorithm.

On my machine running Julia 1.12 I get good performance in almost all the relevant tests. However in the LopezAcosta2019Tiling "sample of cases" and "image types" tests I get segmentation faults, which is weird. Hopefully we'll see soon if that's an issue with Julia 1.11 as well.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,7 +25,7 @@ jobs:
             arch: x64
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       
       - uses: julia-actions/setup-julia@v2
         with:
--- a/.github/workflows/CheckNotebooks.yml
+++ b/.github/workflows/CheckNotebooks.yml
@@ -24,7 +24,7 @@ jobs:
 
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: julia-actions/setup-julia@v2
         with:
--- a/.github/workflows/doc-preview-cleanup.yml
+++ b/.github/workflows/doc-preview-cleanup.yml
@@ -16,7 +16,7 @@ jobs:
       contents: write
     steps:
       - name: Checkout gh-pages branch
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: gh-pages
       - name: Delete preview and history + push changes
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -16,7 +16,7 @@ jobs:
       statuses: write
     runs-on: ubuntu-latest
     steps:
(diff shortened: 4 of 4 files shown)


src/Segmentation/segmented-image-utilities.jl
```diff
@@ -69,3 +71,75 @@ Find pixels in a segmented image with non-zero labels
...
+The algorithm finds all pairs of segment labels at the tile edges. Then, we count the number of 
+times each right-hand label is paired to a left-hand label, and for pairs with at least `minimum_overlap` pixel overlap,
+the right-hand label is assigned as a candidate pair to the left-hand label. If the difference in grayscale
+intensity is less than `grayscale_threshold`, the objects are merged. The function returns an image index map.
+
+"""
+function stitch_clusters(segmented_image, tiles, minimum_overlap=4, grayscale_threshold=0.1) 
+    grayscale_magnitude(c) = Float64(Gray(c))
+    
+    idxmap = deepcopy(segmented_image.image_indexmap)
```
hollandjg: Is this the same as 
```suggestion
    idxmap = deepcopy(labels
_map(segmented_image))
```
?

test/test-segmentation-lopez-acosta-2019-tiling.jl
```diff
@@ -47,7 +47,7 @@ end
     @test 0.62 ≤ round(precision; digits=2)
     @test 0.73 ≤ round(F_score; digits=2)
 end
-
+# Segmentation fault with case 102???
```
hollandjg: Does this happen reliably? Can we split that singe case out as a potentially problematic broken case? Is it only in Julia 1.12?

danielmwatkins: I was getting fairly frequent segmentation faults running on Julia 1.12, but it doesn't seem to be happening with Julia 1.11.

PR #851, opened 2025-12-11
refactor: replace `skimage.exposure.equalize_adapthist` with julia

hollandjg:
Use the Julia `adjust_histogram(..., ContrastLimitedAdaptiveHistogramEqualization(...))` function to replace the similar scikit image function.

In this PR, we only switch out the function; we're not doing any work to remove the casts to and from [0..255] scaled images. That will come later. 

Overall results can be seen in this notebook: https://drive.google.com/file/d/1udu7nm1NM82mze5BrgA9wykwvFzMbNPB/view?usp=sharing

ToDo:
- [x] Run test on 100 validation data set images comparing the two algorithms, and look at what the percentage change is. Include: 128 Hudson Bay (difficult case)

### DIFF
--- a/src/Filtering/clahe.jl
+++ b/src/Filtering/clahe.jl
@@ -186,11 +186,11 @@ function map_histogram(
 ) where {T<:Integer}
     n_pixels = sum(counts)
     scale = (maxval - minval) / n_pixels
-    mapping = similar(counts, Float64) # Can I make this the same type as minval/maxval?
+    mapping = similar(counts, Float64)
     cumulative = 0
     for i in eachindex(counts)
         cumulative += counts[i]
-        mapping[i] = min(minval + cumulative * scale, maxval) 
+        mapping[i] = min(minval + cumulative * scale, maxval)
     end
 
     function _mapping_function(value)


danielmwatkins:
Interesting! This definitely shows a shortcoming of the Julia version. I'd like to see how the algorithms differ under the hood, at the very least looking to see if they are citing similar sources for the algorithm and whether there are any extra filtering steps in the Python one to prevent maxing the gain in dark areas.

danielmwatkins:
OK! I found that Graphics Gems IV is available to checkout from the Internet Archive — you just have to sign in. The intro to the book says that all the code is freely available to use and modify however you wish. It includes C code for the CLAHE algorithm. The original book came with a floppy disk with all the code on it.
https://archive.org/details/graphicsgemsiv0004unse/page/n17/mode/2up 

Take a look and let me know what you think — it might actually be straightforward to port it to Julia.

src/Filtering/clahe.jl
cpaniaguam: I was under the impression this was going to be a [package of its own](https://github.com/brown-ccv/CLAHE.jl/blob/main/src/CLAHE.jl). What's the plan for this?

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    out .= must_resize ? imresize(out_tmp, (height, width)) : out_tmp
+    return out
+end
+
+function validate_parameters(f::ContrastLimitedAdaptiveHistogramEqualization)
+    # !(0 <= f.clip <= 1) && throw(ArgumentError("The parameter `clip` must be in the range [0..1]."))
+    return !(1 <= f.rblocks && 1 <= f.cblocks) &&
+           throw(ArgumentError("At least 1 contextual regions required (1x1 or greater)."))
```
cpaniaguam: The return statement is probably not needed here.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+        @. out_region =
+            (w₁₁ * resultUL + w₁₂ * resultUR + w₂₁ * resultBL + w₂₂ * resultBR) / wₙ
+    end
+
+    out .= must_resize ? imresize(out_tmp, (height, width)) : out_tmp
+    return out
+end
+
+function validate_parameters(f::ContrastLimitedAdaptiveHistogramEqualization)
+    # !(0 <= f.clip <= 1) && throw(ArgumentError("The parameter `clip` must be in the range [0..1]."))
```
cpaniaguam: It feels like this should be kept instead of being commented out. If this is intentional consider deleting it.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    # Size of each contextual region
+    rsize = resized_height / f.rblocks
+    csize = resized_width / f.cblocks
+
+    # Calculate actual clip limit
+    clip_limit = f.clip * (rsize * csize) / f.nbins
+    clip_limit < 1 && (clip_limit = 1)
+    clip_limit = Int(floor(clip_limit))
```
cpaniaguam:
```suggestion
    clip_limit = max(1, floor(Int, f.clip * rsize * csize / f.nbins))
```

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+        img_tmp = imresize(img, (resized_height, resized_width))
+        out_tmp = copy(img_tmp)
+    else
+        img_tmp = img
+        out_tmp = copy(img)
+    end
+
+    # Size of each contextual region
+    rsize = resized_height / f.rblocks
+    csize = resized_width / f.cblocks
```
cpaniaguam: Because of L45-46, one could use `÷` and drop the `Int`-casting downstream.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+end
+
+function redistribute_histogram(
+    counts::AbstractVector{T}, clip_limit::Int
+) where {T<:Integer}
+    n_excess = sum(max(0, count - clip_limit) for count in counts)
+    n_bins = length(counts)
+    if n_excess == 0
+        return counts
+    end
```
cpaniaguam:
```suggestion
    n_excess == 0 && return counts
```

src/Filtering/clahe.jl
hollandjg: Thanks for the feedback! This is super helpful. 
I was working on getting the CLAHE.jl into its own package, and was getting a release ready. As it's not ready to publish yet (you can see from the state of the code that it needs some neatening up), but we're ready to put it into the IceFloeTracker already, I decided to integrate it here first. We'll aim to publish it later in the year, potentially just as a pull request against the ImageSegmentation.jl library.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+"""
+@kwdef struct ContrastLimitedAdaptiveHistogramEqualization{
+    T₁<:Union{Real,AbstractGray},T₂<:Union{Real,AbstractGray},T₃<:Real
+} <: AbstractHistogramAdjustmentAlgorithm
+    nbins::Int = 128
+    minval::T₁ = 0.0
+    maxval::T₂ = 1.0
+    rblocks::Int = 8
+    cblocks::Int = 8
+    clip::T₃ = 0.1
```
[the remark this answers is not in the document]
hollandjg: This is following the format of the `AdaptiveHistogram` algorithm from ImageSegmentation.jl, to reduce friction when we make the pull request.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+        region = view(img_tmp, rstart:rend, cstart:cend)
+        out_region = view(out_tmp, rstart:rend, cstart:cend)
+
+        resultUL, resultUR = histUL.(region), histUR.(region)
+        resultBL, resultBR = histBL.(region), histBR.(region)
+
+        resultUL = histUL.(region)
+        resultUR = histUR.(region)
+        resultBL = histBL.(region)
+        resultBR = histBR.(region)
```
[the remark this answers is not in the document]
hollandjg: Yes, but they shouldn't be. Will deduplicate.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+    n_excess = sum(max(0, count - clip_limit) for count in counts)
+    n_bins = length(counts)
+    if n_excess == 0
+        return counts
+    end
+    increment, remainder = divrem(n_excess, n_bins)
+    new_counts = similar(counts)
+    for i in eachindex(counts)
+        new_counts[i] = min(counts[i], clip_limit) + increment
+    end
```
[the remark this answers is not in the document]
hollandjg: Will take a look at this

src/Filtering/clahe.jl
cpaniaguam: I think your suspicion is correct: the formatter insists on adding a return statement. I personally find it strange to return a short-circuit boolean expression with a side effect. Maybe add `return nothing` at the end?

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,235 @@
...
+        cumulative += counts[i]
+        mapping[i] = minval + cumulative * scale
+        if mapping[i] > maxval
+            mapping[i] = maxval
+        end
+    end
+
+    function _mapping_function(value)
+        index = searchsortedfirst(edges, value)
+        return mapping[index - 1] # -1 because mapped_values is an OffsetArray
```
cpaniaguam: Consider revising this comment. It seems like `mapping` has the same eltype as `counts` (Float64). Also, can `searchsortedfirst` (didn't know Julia had an equivalent of `bisect_left`!) return `1`? If so, this will require special handling.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,235 @@
...
+    #adjust_histogram!(view(yiq_view,1,:,:), f)
+    y = comp1.(yiq)
+    adjust_histogram!(y, f)
+    yiq_view[1, :, :] .= y
+    return out .= convert.(T, yiq)
+end
+
+export ContrastLimitedAdaptiveHistogramEqualization, adjust_histogram, adjust_histogram!
+
+end # module CLAHE
```
cpaniaguam:
```suggestion
end
```

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,235 @@
...
+)
+    validate_parameters(f)
+    height, width = length.(axes(img))
+
+    # If necessary, resize the image so that the requested number of blocks fit exactly.
+    resized_height = ceil(Int, height / (2 * f.rblocks)) * 2 * f.rblocks
+    resized_width = ceil(Int, width / (2 * f.cblocks)) * 2 * f.cblocks
+    must_resize = (resized_height != height) || (resized_width != width)
+    if must_resize
+        img_tmp = imresize(img, (resized_height, resized_width))
```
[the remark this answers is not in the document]
hollandjg: I agree that it's a good idea to pad rather than resize. I'll try to get reflection padding working – that seems appropriate.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,237 @@
...
+    clip_limit < 1 && (clip_limit = 1)
+    clip_limit = Int(floor(clip_limit))
+
+    # Process each contextual region
+    histograms = Array{Any}(undef, f.rblocks, f.cblocks)
+    for rblock in 1:(f.rblocks), cblock in 1:(f.cblocks)
+        rstart = Int((rblock - 1) * rsize) + 1
+        rend = Int(rblock * rsize)
+        cstart = Int((cblock - 1) * csize) + 1
+        cend = Int(cblock * csize)
```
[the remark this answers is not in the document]
hollandjg: Yes. I didn't want to implement the bilinear interpolation as they'd written it in the original paper (it was unnecessarily low-level for this high-level language) so I used some equations from Wikipedia. To avoid getting confused, I kept the Wikipedia symbols. They can and probably should be unified.

src/Filtering/clahe.jl
hollandjg: Will take a look. Sounds very sensible to me.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,235 @@
...
+        x₁, x₂ = rstart, rend
+        y₁, y₂ = cstart, cend
+        x = Array(range(rstart, rend))
+        y = Array(range(cstart, cend))'
+
+        w₁₁ = ((x₂ .- x) .* (y₂ .- y))
+        w₁₂ = ((x₂ .- x) .* (y .- y₁))
+        w₂₁ = ((x .- x₁) .* (y₂ .- y))
+        w₂₂ = ((x .- x₁) .* (y .- y₁))
+        wₙ = ((x₂ - x₁) * (y₂ - y₁))
```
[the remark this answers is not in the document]
hollandjg: Yes.

src/Filtering/clahe.jl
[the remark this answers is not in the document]
hollandjg: Updated to use padding and reflection on the boundaries

src/Filtering/clahe.jl
hollandjg: Done

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,219 @@
...
+    # Size of each contextual region
+    rsize = resized_height ÷ f.rblocks
+    csize = resized_width ÷ f.cblocks
+
+    # Calculate actual clip limit
+    clip_limit = f.clip * (rsize * csize) / f.nbins
+    clip_limit < 1 && (clip_limit = 1)
+    clip_limit = Int(floor(clip_limit))
+
+    # Process each contextual region
```
cpaniaguam: Consider adding a reference to #869.

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,219 @@
...
+
+        # Interpolate the results using the values and the weights
+        vₒ = @. (w₁₁ * v₁₁ + w₁₂ * v₁₂ + w₂₁ * v₂₁ + w₂₂ * v₂₂) / wₙ
+
+        # Write to the correct region in the output image
+        @. out_region = vₒ
+    end
+
+    out .= must_pad ? out_padded[1:height, 1:width] : out_padded
+    return out
```
cpaniaguam: I can you can get away with just `out_padded[1:height, 1:width]` regardless.

```suggestion
    return out_padded[1:height, 1:width]
```

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,219 @@
...
+
+        # Get the weights of each value within the block
+        w₁₁ = @. ((rₑ - r) * (cₑ - c))
+        w₁₂ = @. ((rₑ - r) * (c - cₛ))
+        w₂₁ = @. ((r - rₛ) * (cₑ - c))
+        w₂₂ = @. ((r - rₛ) * (c - cₛ))
+        wₙ = ((rₑ - rₛ) * (cₑ - cₛ))
+
+        # Interpolate the results using the values and the weights
+        vₒ = @. (w₁₁ * v₁₁ + w₁₂ * v₁₂ + w₂₁ * v₂₁ + w₂₂ * v₂₂) / wₙ
```
danielmwatkins: Love the clear notation here!

src/Filtering/clahe.jl
```diff
@@ -0,0 +1,219 @@
...
+    out::GenericGrayImage, img::GenericGrayImage
+)
+    validate_parameters(f)
+    height, width = length.(axes(img))
+
+    # If necessary, resize the image so that the requested number of blocks fit exactly.
+    resized_height = ceil(Int, height / (2 * f.rblocks)) * 2 * f.rblocks
+    resized_width = ceil(Int, width / (2 * f.cblocks)) * 2 * f.cblocks
+    must_pad = (resized_height != height) || (resized_width != width)
+    if must_pad
```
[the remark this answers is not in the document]
hollandjg: I'll add a `@debug`

Issue #820, opened 2025-11-07
Feat: Julia implementation of regionprops_table


danielmwatkins:
The ImageMorphology library contains a handful of `component_*` functions that can provide the measurements that we are looking for. The key properties that we need for tracking, and corresponding ImageMorphology functions, are:
* label 
* area: `component_length` or `segment_pixel_count`
* convex area: `convexhull()` (produces cartesian indices, not boolean image)
* centroid: `component_centroids`
* bounding box: `component_boxes`
* major axis
* minor axis
* orientation

The existing code already takes the step to reframe the Python centroid-0 and centroid-1 ouputs into row_centroid and col_centroid. Additionally it would help to have the average colors. That's not implemented in regionprops.jl yet, but is easily accessible through `segment_means().`

The major axis, minor axis, and orientation can be calculated from the moments of each shape. (see e.g. section 8.6.1 of Digital Image Processing by Burger and Burge. 

For the convex area, we can use Green's theorem and calculate the sum
$$ A = \sum_{k} (x_k y_{k+1} - x_{k+1}y_k $$
where the convex hull is defined by the line (x_i, y_i). I'll just need to check whether the convex hull output fully wraps around the object or if we have to append the first index, and if the output of convexhull() is in order.

PR #854, opened 2025-12-18
Feat julia regionprops

danielmwatkins:
Addresses issue #820 

Tasks:
- [x] area
- [x] convex area
- [x] perimeter
- [x] centroid
- [x] bbox
- [x] ellipse (maj/min, orientation)
- [ ] latlon conversion
- [x] masks
- [ ] others? (psi-s, perhaps)
- [x] split up regionprops and regionprops table, since the table function is just sending the regionprops dict to DataFrame

Done so far: 

`regionprops_table` function with basic structure, new functions for component floe masks and an implementation of the Benkrid-Crookes perimeter algorithm.

### DIFF
--- a/src/Segmentation/regionprops.jl
+++ b/src/Segmentation/regionprops.jl
@@ -346,8 +346,8 @@ function (f::PixelConvexArea)(A)
             
         chull = convexhull(A .== i)
         N = length(chull)
-        x = getindex.(collect(bboxes[i]), 1)
-        y = getindex.(collect(bboxes[i]), 2)
+        x = getindex.(bboxes[i], 1)
+        y = getindex.(bboxes[i], 2)
         
         for idx in eachindex(x)
             xi, yi = x[idx], y[idx]


danielmwatkins:
I've added a basic computation for the convex hull which uses the convex point-in-polygon algorithm. This doesn't end up being exactly the same as the Python version, but the results are close. I also included the method computing the area of the convex hull polygon directly with Green's theorem. The scatter plot shows the new method on the X-axis and the Python method on the Y-axis. You can see there is some underestimation. However, since both are approximations, I'm not sure it matters a whole lot. Either way, I think that the functions I put together could be sped up but at least it works and its in Julia.

image

I edited the tests because it had been including a 1-pixel region, which is too small to compute a convex hull. We could add a default of 1, but I think it's better to just exclude such cases.

danielmwatkins:
Another thing still to do -- there's the regionprops as well as the regionprops table functions. What we might want to do is change the name of my function to regionprops and remove the DataFrame line at the end, then have a simple wrapper function that runs regionprops and sends it to the dataframe.

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+    algorithm_functions = Dict("benkrid_crookes" => benkrid_crookes)
+
+    for label in keys(masks)
+        n, m = size(masks[label])
+        n * m == 1 && continue
+        label == 0 && (masks[label] = 0; continue)
+        
+        # Shape needs to have a border of zeros for erode to work here    
+        mpad = padarray(masks[label], Fill(0, (1, 1)))
+        connectivity == 4 ? (strel = strel_diamond((3,3))) : (strel = strel_box((3,3)))
```
cpaniaguam: If connectivity doesn't depend on `label`, it can be pulled out of the loop. And one could choose the type of strel first and then pass the size to make the statement a bit more DRY.

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+    type_vals[[21, 33]] .= sqrt(2)
+    type_vals[[13, 23]] .= (1 + sqrt(2)) / 2
+
+    conv_arr = [10 2 10; 2 1 2; 10 2 10]
+    results = conv(edge_array, conv_arr; algorithm=:direct)
+
+    val_counts = Dict()
+    for val in vec(results)
+        val ∉ keys(val_counts) && (val_counts[val] = 0)
+        val_counts[val] += 1
```
cpaniaguam: Here one could use the `get` function with default value, akin to the `.get` method for Python dicts.
```suggestion
        val_counts[val] = get(val_counts, val, 0) + 1
```

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+  Row │ area   perimeter 
+      │ Int32  Float64   
+ ─────┼──────────────────
+    1 │    13   11.6213
+    2 │     1    0.0
+    3 │     1    0.0
+    4 │     7    4.62132
+```
+"""
+function regionprops_table(
```
cpaniaguam: There is a lot happening in this function. I would split it up into functional pieces, one for each computed prop.

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+
+Estimate the perimeter of the labeled regions in `indexmap` using the specified algorithm.
+Algorithm options = "benkrid_crookes" (only option currently, will add crofton in future release)
+
+
+"""
+function component_perimeters(
+    indexmap;
+    minimum_area=1, 
+    algorithm="benkrid_crookes",
```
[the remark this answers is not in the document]
danielmwatkins: Yes, I think that would be more consistent with the rest of the code.

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+"""component_perimeter(indexmap; minimum_area=1, algorithm="benkrid_crookes")
+
+Estimate the perimeter of the labeled regions in `indexmap` using the specified algorithm.
+Algorithm options = "benkrid_crookes" (only option currently, will add crofton in future release)
+
+
+"""
+function component_perimeters(
+    indexmap;
+    minimum_area=1,
```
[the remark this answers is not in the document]
danielmwatkins: That's true -- I'm not sure what happens in the Benkrid Crookes method for a 1-pixel perimeter. From prior tests I know the error in perimeter gets quite large as the number of pixels goes to zero.

src/Segmentation/regionprops.jl
danielmwatkins: Oh true, I'll do that. What does "DRY" mean in this case?

cpaniaguam: "Don't Repeat Yourself"

src/Segmentation/regionprops.jl
danielmwatkins: I've done that for most of the measures. The ones that are longer are ones where it uses previous measures to calculate new ones, like with computing image moment-based measures like major and minor axis length. What I'd need to figure out there is how to get the information that's needed for each measure without recalculating it each time.

danielmwatkins:
Updating the pixel method I found some errors in the logic, luckily, so I fixed those. Here's an example of the convex area, testing the updated logic: first showing the floe, second showing the pixels classified as inside the convex hull.

floe_and_convex_hull

convex_floe_and_convex_hull

src/Segmentation/regionprops.jl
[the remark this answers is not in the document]
danielmwatkins: I've updated it to use the functor approach for the subprocess algorithms.

src/Segmentation/regionprops.jl
```diff
@@ -297,3 +311,413 @@ function converttounits!(propdf, latlondata, colstodrop)
...
+
+
+"""component_convex_area(A; method="pixels")
+
+Compute the convex area of labeled regions. Two methods available: "pixel" and "polygon".
+The polygon method uses Green's theorem to find the area of a polygon through its line integral, 
+while the pixel method uses a point-in-pixel calculation to determine if pixels are inside the
+convex hull. In general the polygon area will be smaller than the pixel area.
+"""
+function component_convex_area(A::AbstractArray{T}; method="pixel") where {T<:Integer}
```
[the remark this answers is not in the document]
danielmwatkins: Done.

src/skimage/skimage.jl
```diff
@@ -2,15 +2,16 @@ module skimage
 
 using PyCall
 
-export sk_measure, sk_morphology, sk_exposure
+# export sk_measure, sk_morphology, sk_exposure
```
cpaniaguam:
```suggestion
```

test/test-add_floemasks.jl
```diff
@@ -10,8 +10,8 @@
             0 0 3 0 4 4 4
         ]
         seg_img = SegmentedImage(labels, labels)
-        props = regionprops_table(seg_img)
-        add_floemasks!(props, seg_img)
+        props = regionprops_table(seg_img; properties=[:label, :mask])
+        # add_floemasks!(props, seg_img)
```
cpaniaguam:
```suggestion
```

src/Segmentation/regionprops.jl
```diff
@@ -158,142 +482,176 @@ julia> bw_img = rand([0, 1], 5, 10)
...
+        mu02 = _central_moment(X, Y, xc, yc, 0, 2)
+        theta = 0.5 * atan(2*mu11, mu20 - mu02)
+        
+        λ0 = (mu20 + mu02 + sqrt((mu20 - mu02)^2 + 4*mu11^2))/2
+        λ1 = (mu20 + mu02 - sqrt((mu20 - mu02)^2 + 4*mu11^2))/2
+        
+        ## 
+        ra = 4*(λ0/areas[s])^0.5 
+        rb = 4*(λ1/areas[s])^0.5
+        append!(moment_measures, [[ra, rb, theta]])
```
cpaniaguam: I like the use of `λ`. Could `θ` and `μ` be used for `theta` and `mu` variable names?

src/Segmentation/regionprops.jl
```diff
@@ -158,142 +482,176 @@ julia> bw_img = rand([0, 1], 5, 10)
...
+        
+        mu11 = _central_moment(X, Y, xc, yc, 1, 1)
+        mu20 = _central_moment(X, Y, xc, yc, 2, 0)
+        mu02 = _central_moment(X, Y, xc, yc, 0, 2)
+        theta = 0.5 * atan(2*mu11, mu20 - mu02)
+        
+        λ0 = (mu20 + mu02 + sqrt((mu20 - mu02)^2 + 4*mu11^2))/2
+        λ1 = (mu20 + mu02 - sqrt((mu20 - mu02)^2 + 4*mu11^2))/2
+        
+        ##
```
cpaniaguam:
```suggestion
```

test/test-regionprops.jl
```diff
@@ -44,28 +44,23 @@
...
             length(unique(label_components(table.mask[i], trues(3, 3)))) for
             i in 1:nrow(table)
-        ] .== [2, 2, 1],
+        ] .== [2, 2],
     )
+
+    # Test that algorithm options at least run
+    _ = regionprops(label_img; properties=["perimeter"], perimeter_algorithm=BenkridCrookes())
+    _ = regionprops(label_img; properties=["convex_area"], convex_area_algorithm=PolygonConvexArea())
+    _ = regionprops(label_img; properties=["convex_area"], convex_area_algorithm=PixelConvexArea())
```
cpaniaguam:
```suggestion
    # Test that algorithm options at least run
    regionprops(label_img; properties=["perimeter"], perimeter_algorithm=BenkridCrookes())
    regionprops(label_img; properties=["convex_area"], convex_area_algorithm=PolygonConvexArea())
    regionprops(label_img; properties=["convex_area"], convex_area_algorithm=PixelConvexArea())
```

test/test-regionprops.jl
```diff
@@ -44,28 +44,23 @@
...
-    # check default properties
-    @test table == regionprops_table(label_img)
-
     # Tests for regionprops
     regions = regionprops(label_img, bw_img)
-
-    # Check some data matches in table
-    randnum = rand(1:total_labels)
-    @test table.area[randnum] == regions[randnum].area
+    print(keys(regions))
```
cpaniaguam: Is this needed? If so, consider using `@info` instead which provides better logging control.

src/Segmentation/regionprops.jl
```diff
@@ -158,142 +482,176 @@ julia> bw_img = rand([0, 1], 5, 10)
...
-    props[:, props_to_fix] .+= 1
-    return nothing
+    # psi-s needs masks, so this can get called first
+    :mask ∈ properties && begin
+        floe_masks = component_floes(labels; minimum_area=minimum_area)
+        push!(data, :mask => map(s -> floe_masks[s], img_labels))
+    end
+
+    # TODO add psi-s curve generation as an option
+    updated_properties = []
```
cpaniaguam:
```suggestion
    updated_properties = Symbol[]
```

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,376 @@
...
+while the pixel method uses a point-in-pixel calculation to determine if pixels are inside the
+convex hull. In general the polygon area will be smaller than the pixel area.
+"""
+function component_convex_area(A;
+    algorithm::ConvexAreaEstimationAlgorithm=PixelConvexArea()
+    ) 
+    mn, mx = extrema(A)
+    if !(mn == 0 || mn == 1)
+        throw(ArgumentError("The input labeled array should contain background label `0` as the minimum value"))
+    end
```
cpaniaguam: Consider moving this validation routine to a helper function and reuse here and in the `component_floes` function (and elsewhere!).

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,376 @@
...
+end
+
+
+"""PixelConvexArea()
+
+Estimate the convex area by integrating the area of the convex hull polygon.
+Uses the convexhull function from ImageMorphology, which raises an error if the
+area of the segment is less than or equal to 3. In general, the error should be smaller
+for larger shapes.
+"""
```
cpaniaguam: The docsting's content says the same as that of `PolygonConvexArea`. Is this intentional?

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,376 @@
...
+@kwdef struct PixelConvexArea <: ConvexAreaEstimationAlgorithm
+    minimum_area = 4
+end
+
+function (f::PixelConvexArea)(A)
+    _, mx = extrema(A)
```
cpaniaguam: If only the max is need consider using `maximum`.
```suggestion
    mx = maximum(A)
```

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,376 @@
...
+Uses the convexhull function from ImageMorphology, which raises an error if the
+area of the segment is less than or equal to 3. In general, the error should be smaller
+for larger shapes.
+"""
+@kwdef struct PolygonConvexArea <: ConvexAreaEstimationAlgorithm
+    minimum_area = 4
+end
+
+function (f::PolygonConvexArea)(A)
+    mn, mx = extrema(A)
```
cpaniaguam:
```suggestion
    mx = maximum(A)
```

src/Segmentation/regionprops.jl
```diff
@@ -146,10 +470,10 @@ julia> bw_img = rand([0, 1], 5, 10)
   0  1  0  1  0  0  0  0  4  0
   1  0  0  0  0  2  0  4  0  4
 
- julia> regions = regionprops(label_img, bw_img);
+ julia> measures = regionprops(label_img, bw_img);
 
- julia> for region in regions
-           println(region.area,"\t", region.perimeter)
+ julia> for s in unique(label_img)
+           println(region.area[s],"\t", region.perimeter[s])
```
cpaniaguam: Should this be using `measures` instead of `region`?
```suggestion
           println(measures.area[s],"\t", measures.perimeter[s])
```

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,376 @@
...
+Compute measures of labeled regions in label_img and return as a DataFrame. Optionally, include an
+extra image or array associated with the labels.
+        
 # Arguments
-- `label_img`: Image with the labeled objects of interest
-- `intensity_img`: (Optional) Used for generating `extra_properties`, integer/float array from which (presumably) `label_img` was generated 
+- `label_img`: Image with the labeled objects of interest. May be an integer array or a SegmentedImage.
+- `intensity_img`: (Optional) Used for generating `extra_properties`, such as a color image to use for calculating mean color in segments.
 - `properties`: List (`Vector` or `Tuple`) of properties to be generated for each connected component in `label_img`
 - `extra_properties`: (Optional) not yet implemented. It will be set to `nothing`
```
cpaniaguam: Consider adding a word or two about the new arguments `minimum_area`, `perimeter_algorithm`, and `convex_area_algorithm`.

cpaniaguam (changes requested):
This is looking great! I added a few more suggestions. Consider applying formatting to the files as I noted some inconsistent indentation instances.

src/Segmentation/regionprops.jl
danielmwatkins: Fair, this was mirroring the other `component_*` functions from image morphology and used the minimum for checking if the background was 0. I dropped the mn since it wasn't used in that section.

src/Segmentation/regionprops.jl
```diff
@@ -158,142 +482,176 @@ julia> bw_img = rand([0, 1], 5, 10)
...
     extra_properties::Union{Tuple{Function,Vararg{Function}},Nothing}=nothing,
+    minimum_area=1,
+    perimeter_algorithm=BenkridCrookes(),
+    convex_area_algorithm=PixelConvexArea()
 )
-    if !isnothing(extra_properties)
-        @error "<extra_properties> not yet implemented in this wrapper; setting it to <nothing>"
-        extra_properties = nothing
+
+    isa(label_img, SegmentedImage) ? (labels = labels_map(label_img)) : labels = deepcopy(label_img)
```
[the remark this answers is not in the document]
danielmwatkins: Probably not actually

cpaniaguam:
> @cpaniaguam I've cleaned up some of the formatting, let me know if I didn't catch the things that you had in mind.

Thanks @danielmwatkins. There are a few _unresolved_ discussions I'm not sure if you looked  at them. If you did, consider marking them as _resolved_.

danielmwatkins:
@cpaniaguam can you be more specific? I don’t see any of them. All the comments and suggestions I see are resolved.

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,390 @@
...
+    perims = Dict()
+
+    for label in keys(masks)
+        n, m = size(masks[label])
+        n * m == 1 && continue
+        label == 0 && (perims[label] = 0; continue)
+        
+        # Shape needs to have a border of zeros for erode to work here    
+        perims[label] = algorithm(masks[label])
+    end
```
cpaniaguam: Would this work here?

```suggestion
    perims = Dict(
        label => (label == 0 ? 0 : algorithm(masks[label])) for
        label in keys(masks) if prod(size(masks[label])) > 1
    )
```

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,390 @@
...
+used for the erosion is the only parameter. The algorithm uses strel_diamond((3,3)) for 4-connectivity and
+strel_box((3,3)) for 8-connectivity. The resulting function operates on a binary array, which 
+is assumed to contain a single object.
+
+# Examples
+```
+julia> benkrid_crookes = BenkridCrookes(connectivity=4)
+julia> A = [1 1 1; 1 1 1; 1 1 1]
+julia> benkrid_crookes(A)
+8.0
```
cpaniaguam: I think this sample array is a bit more interesting in that the two algos produce a different result.
```suggestion
julia> A = [0 1 1; 1 1 1; 1 1 1];

julia> BenkridCrookes(connectivity=4)(A)
7.414213562373095

julia> BenkridCrookes(connectivity=8)(A)
7.0
```

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,390 @@
...
+        end
+            
+        chull = convexhull(A .== i)
+        N = length(chull)
+        x = getindex.(collect(bboxes[i]), 1)
+        y = getindex.(collect(bboxes[i]), 2)
+        
+        for idx in 1:length(x)
+            xi = x[idx]
+            yi = y[idx]
```
cpaniaguam:
```suggestion
        for idx in eachindex(x)
            xi, yi = x[idx], y[idx]
```

src/Segmentation/regionprops.jl
```diff
@@ -158,142 +507,174 @@ julia> bw_img = rand([0, 1], 5, 10)
...
+
+    row_centroid = first.(centroids)
+    col_centroid = last.(centroids)
+    push!(data, :row_centroid => map(s -> row_centroid[s], label_list))
+    push!(data, :col_centroid => map(s -> col_centroid[s], label_list))
+
+    moment_measures = []
+    for s in label_list
+        X = getindex.(indices[s], 1)
+        Y = getindex.(indices[s], 2);
```
cpaniaguam: I don't see a call to `collect` here hence my hunch that it might not be necessary above.

cpaniaguam (approved):
This is great, thanks @danielmwatkins! Left a few suggestions for your consideration.

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,390 @@
...
+    return perim
+end
+
+
+"""component_convex_area(A; algorithm=PixelConvexArea()")
+
+Compute the convex area of labeled regions. Two methods available: "PixelConvexArea" and "PolygonConvexArea".
+The polygon method uses Green's theorem to find the area of a polygon through its line integral, 
+while the pixel method uses a point-in-pixel calculation to determine if pixels are inside the
+convex hull. In general the polygon area will be smaller than the pixel area.
```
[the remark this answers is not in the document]
danielmwatkins: I’m not sure, I’ll have to see if this is named in one of the image processing books I’ve been using. Or see if I can prove it. It’s empirically true for the ice floe data I was testing it with, but it may depend on there being some regularity to the floe boundary.

src/Segmentation/regionprops.jl
```diff
@@ -2,25 +2,390 @@
...
+        # treat convex area background and too-small objects as undefined
+        (i == 0) || (areas[i] < f.minimum_area) && begin
+            convex_areas[i] = NaN
+            continue
+        end
+            
+        chull = convexhull(A .== i)
+        N = length(chull)
+        x = getindex.(collect(bboxes[i]), 1)
+        y = getindex.(collect(bboxes[i]), 2)
```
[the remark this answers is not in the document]
danielmwatkins: I still have to do a lot of trial and error to go in between cartesianindex objects and position vectors. I think I was getting errors if I didn’t collect first.

PR #861, opened 2026-01-21
feat: add snakemake workflow

hollandjg:


hollandjg:
Tightly coupled with the shape of the CLI. Merging into the CLI PR.

PR #863, opened 2026-01-27
add tiled kmeans

danielmwatkins:
This extends the k-means method to run on a provided tiled iterator. It uses the stitch clusters function to align objects across edges.


danielmwatkins:
Further updates
* Added explicit tests for the kmeans functions (segmentation, binarization, both tiled and non-tiled)
* Added helper functions for viewing segmentation results
* Added test for helper functions
* Updated docstrings for kmeans functions

src/Segmentation/k-means-segmentation.jl
cpaniaguam: Consider applying the formatter to this file. There are few lines that seem too long.

PR #865, opened 2026-02-07
Switch to more flexible imgradient function

danielmwatkins:
The imgradientmag function was using a non-normalized Sobel kernel, which was hard-coded. This PR replaces that with the imgradient function from ImageFiltering, allowing the user to specify a different kernel if desired.

### DIFF
--- a/test/test-img-processing.jl
+++ b/test/test-img-processing.jl
@@ -11,7 +11,7 @@
 
     @testset "imgradientmag" begin
         gmag = to_uint8(imgradientmag(coins))
-        @test sum(gmag) == 2938959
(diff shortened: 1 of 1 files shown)


src/Filtering/imgradientmag.jl
```diff
@@ -1,15 +1,12 @@
-import Images: centered
-
+import Images: imgradients
+import Images.KernelFactors: sobel
 """
     imgradientmag(img)
```
cpaniaguam: Update this to reflect new arg

cpaniaguam (changes requested):
Looks good. Just one little thing.

PR #871, opened 2026-03-06
Chore: Clean up Tracker module 

danielmwatkins:
The updated FloeTracker functor replaced the two former tracker methods, merging ideas and methods from both into a more flexible structure. This PR focuses on removing unused functions, updating tests to touch functions that we still want to keep, and consolidating functions where appropriate.


test/test-floe-tracker.jl
cpaniaguam: Hit with formatter.

PR #879, opened 2026-03-17
Update segmented image viewer

danielmwatkins:
I made two changes to the segmented image viewers:
1. Automatically set the background to black
2. Automatically choose a new color if the color is too dark


cpaniaguam (approved):
👍

PR #903, opened 2026-04-01
Add truecolor peaks ice detection function

danielmwatkins:
This PR adds a simple functor and an associated test. The functor is an ice detection algorithm which uses band 1 of the truecolor image to identify sea ice pixels.

image

Above: Example truecolor images.

The method assumes the land mask and cloud mask has already been applied, and just takes in the minimum band 1 reflectance, the window size for `get_ice_peaks`, and the minimum prominence for `get_ice_peaks`. In the case of no distinct peak, it just uses the `band_1_min` as the threshold. The example below shows the mask produced after the Watkins2025CloudMask is applied, and varying band 1 min from 0.1 to 0.5.

image


hollandjg (approved):
Very neat!

PR #909, opened 2026-04-03
add function to load landfast ice

danielmwatkins:
Adds a loader tool to access the landfast ice mask from a case.


danielmwatkins:
@hollandjg easy when the template is right there!

PR #932, opened 2026-04-09
refactor: move binarize_mask to ImageUtils

hollandjg:
Provide a `binarize_mask` function required for #870 which can be used everywhere. 

Before this change, `binarize_landmask` couldn't be used in the Data module because Preprocessing is imported later than Data. Since `binarize_landmask` has no internal dependencies, it should be defined in one of the earliest modules.


src/ImageUtils/mask.jl
```diff
@@ -200,3 +200,22 @@ end
...
+    binarize_mask(mask_image)
+    binarize_mask(mask_image; tol=0.1)
+
+Convert a 3-channel RGB or 1-channel Gray mask image to a 1-channel binary matrix with mask = 1, everything else = 0.
+Assumes that the input image is 0 over the parts to be left unmasked, and some shade over the parts to be masked. 
+The tol argument lets a higher threshold for masked pixels be chosen.
+
+# Arguments
+- `mask_image`: mask image, e.g. land mask from NASA Worldview
+- `tol` (Optional): Values in the image larger than `tol` are considered land.
```
danielmwatkins: For generality, how about values larger than `tol` are set to 1? Since this could be used for other masks than just land masks.

src/Preprocessing/landmask.jl
```diff
@@ -27,7 +28,7 @@ function create_landmask(
     fill_value_lower::Int=0,
     fill_value_upper::Int=2000,
 ) where {T<:AbstractMatrix}
-    land_mask = binarize_landmask(landmask_image)
+    land_mask = binarize_mask(landmask_image)
```
danielmwatkins, line 36: I'd still like to keep the `create_landmask` function. The idea being we can use binarize mask internally when it's an image, but later add a method to take in a coast shapefile.

src/ImageUtils/mask.jl
hollandjg: Yes!

src/Preprocessing/landmask.jl
hollandjg: Yes! We might want to refactor `create_landmask` later to match the shape of `create_coastal_buffer_mask` (just return one output). This PR shouldn't change `create_landmask`'s function at all.

danielmwatkins: Good idea!

danielmwatkins (approved):
Looks good to me!

PR #937, opened 2026-04-16
feat: reorganize files by filetype

hollandjg:
Add rules to reorganize files by type.

### DIFF
--- a/workflow/README.md
+++ b/workflow/README.md
@@ -112,6 +112,22 @@ Each row specifies:
 - the image scale in metres
 - the segmentation algorithm type.
 
+To reorganize the output files by filetype, use the `region_case_results_by_filetype` target: 
+```shell
+snakemake -c 4 region_case_results_by_filetype
+```
+
+which links each output file into a new subdirectory organized by filetype and, where appropriate, pipeline, 
+e.g.:
+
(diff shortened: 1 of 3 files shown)


hollandjg:
@copilot wherever there is a risk of a failing link or file copy, make it a safe overwrite, e.g. using `ln --force`.

workflow/README.md
```diff
@@ -101,16 +101,32 @@ Each region is identified by a name, and specified by:
 ## Batch processing
 
 To run the tracking pipeline on a series of time periods and regions, we can run:
-```
-snakemake -c 4 region-case-results.txt
+```shell
+snakemake -c 4 region_case_results
```
danielmwatkins: Does this parse the order like region, case, results, or what does the name here signify? Or is "region_case_results" a recognized keyword command?

danielmwatkins: I see that "region_case_results" is the name of a rule. It might help to say something like, we use the syntax snakemake -c <number of cores> <name of rule> and support the following options for rules (or point to where rules are defined).

hollandjg: Yes! I made an update to the readme.

danielmwatkins (approved):
Updates to the README look good, thanks!

PR #939, opened 2026-04-20
Update test cases to v0.2

danielmwatkins:
* Update Watkins2026DataSet references to v0.2
* Change the check on the labeled and binary image loader from "fl_analyst" to "number_floes"


danielmwatkins:
Slightly lower recall on the "detailed" case for the tiling algorithm -- the v0.2 update includes more floes due to including the objects intersecting the floe boundary, so the denominator in the recall equation gets larger.

cpaniaguam (approved):
Lgtm

PR #944, opened 2026-04-23
Fix  shape difference inexact error

hollandjg:


hollandjg:
I'm not sure this is the right way to go. Will revisit this PR later when we've solved https://github.com/WilhelmusLab/IceFloeTracker.jl/issues/913.

PR #945, opened 2026-04-23
Fix: missing columns in dataframes

hollandjg:
- Removes weird `Val{:raw}` syntax in filter functions
- resolves 913


hollandjg:
I'm not sure this is the right way to go. I think `DataFrames.transform` and `DataFrames.subset` would be better. 
Superseded by #949

PR #946, opened 2026-04-23
always include columns from filter function even if there are no rows

hollandjg:
- part of solution for #913


hollandjg:
This doesn't actually resolve the problem.

PR #947, opened 2026-04-23
refactor: simplify filter function syntax

hollandjg:
- Fixes a long-standing wyrdness in the definition of the filter function by having them always be called using `map` or `filter`
- part of the resolution for #913


hollandjg:
No. Superseded by #949

PR #949, opened 2026-04-28
Implement DataFrames transform and subset syntax for candidate filtering

hollandjg:


hollandjg:
This is more like it, but it breaks our functional approach, and means we always have to use the weird DataFrames transform syntax. Is there a better way to go?

PR #951, opened 2026-04-30
Fix error in masking in example notebook

danielmwatkins:
- resolves #918


hollandjg:
- resolves #918

hollandjg (approved):
Looks good to me.

PR #955, opened 2026-05-01
fix: minimum weight matching function by adding column existence check

hollandjg:
Add a column existence check at the start of the MinimumWeightMatchingFunction.

- resolves #913

### DIFF
--- a/src/Tracking/matching_functions.jl
+++ b/src/Tracking/matching_functions.jl
@@ -33,7 +33,7 @@ the forward and backward grouped minimizations are identified as likely true mat
 end
 
 function (f::MinimumWeightMatchingFunction)(candidate_pairs::DataFrame)
-    # djw, jgh: Ideally, we woulnd't need this guard, it's not as clean as it could be, but is okay for now.
+    # dmw, jgh: Ideally, we woulnd't need this guard, it's not as clean as it could be, but is okay for now.
     if !(String.(f.columns) ⊆ names(candidate_pairs))
         @debug(
             "Columns specified in MinimumWeightMatchingFunction not found in candidate_pairs DataFrame. Returning an empty dataframe.",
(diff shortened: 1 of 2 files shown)


src/Tracking/matching_functions.jl
```diff
@@ -33,6 +33,13 @@ the forward and backward grouped minimizations are identified as likely true mat
 end
 
 function (f::MinimumWeightMatchingFunction)(candidate_pairs::DataFrame)
+    # djw, jgh: Ideally, we woulnd't need this guard, it's not as clean as it could be, but is okay for now.
```
danielmwatkins: dmw, not djw (unless you're talking about someone else)

src/Tracking/matching_functions.jl
```diff
@@ -49,9 +56,10 @@ function (f::MinimumWeightMatchingFunction)(candidate_pairs::DataFrame)
         sdf -> sdf[argmin(sdf.w), :], groupby(candidate_pairs, :head_uuid)
     )
 
-    return innerjoin(
+    result = innerjoin(
         matches_fwd[:, [:head_uuid, :uuid]], matches_bwd; on=[:head_uuid, :uuid]
     )
+    return result
```
danielmwatkins: What is the benefit of creating the "result" object rather than returning it directly?

src/Tracking/matching_functions.jl
```diff
@@ -33,6 +33,13 @@ the forward and backward grouped minimizations are identified as likely true mat
 end
 
 function (f::MinimumWeightMatchingFunction)(candidate_pairs::DataFrame)
+    # djw, jgh: Ideally, we woulnd't need this guard, it's not as clean as it could be, but is okay for now.
+    if !(String.(f.columns) ⊆ names(candidate_pairs))
```
danielmwatkins, line 37: Nice use of the subset notation!

src/Tracking/matching_functions.jl
hollandjg:
```suggestion
    # dmw, jgh: Ideally, we woulnd't need this guard, it's not as clean as it could be, but is okay for now.
```

hollandjg: Sorry, got you name wrong!

danielmwatkins (approved):
Looks good to me!

PR #960, opened 2026-05-07
fix: add support for antarctica in the data loader

hollandjg:
- resolves #956

- [x] Update reference to loader once https://github.com/brown-ccv/earthdata-downloader/pull/2 is merged


danielmwatkins (approved):
Update looks good to me! Thanks for fixing this.

PR #967, opened 2026-05-11
Streamlined pipeline for cal-val

danielmwatkins:
The FSPipeline is a new pipeline using both semantic and object-based segmentation approaches. It includes two main methods for producing binary sea ice floe masks: the $k$-means method, used in the prior versions, and an adaptive threshold method, similar to what is used in the `ebseg` method. 

The pipeline produces a cloud mask using the Watkins2026 method, an ice/water mask based on the IceDetectionBrightnessMidpoint method, and uses an explicit Preprocess functor to perform the sharpening and equalization steps.

After the binarization, the images are lightly cleaned and smoothed using morphological operations and conditional hole filling. Holes in the binary mask that either contain sea ice or cloud are filled up to a set size. 

Next, we use a new floe splitting procedure, `dist_morph_split`, which applies a series of thresholds to the distance function up to a maximum depth, then labels distinct objects and re-grows them. 

We include a dedicated floe filter function which clears objects which are (a) outside the specified size range, (b) intersect the coastal buffer mask, (c) are not circular enough, or (d) are too dark in Band 2 of the falsecolor image. The settings for floe size, color, and circularity are specified separately based on the object-average cloud fraction, so cloud-covered objects can be filtered more strictly than clear-sky if desired.

Finally, we use a `merge_floes` function to compare the results of the two segmentation methods. This step resolves collisions between the two candidate floe labels. For each floe, we identify a "relevant set" based on the location of object centroids and the degree of overlap. Then, if the objects agree within a distance and area based threshold, the more circular object is taken. For all other objects, collisions are resolved based on the object prominence, calculated as the difference between the object mean brightness and the brightness of the surroundings. This calculation is accelerated by using the expand labels function -- hence, if floes are tightly packed, the expanded boundary area may be truncated compared to an object that is well-separated from neighboring floes.


fspipeline_test.jl
cpaniaguam, line 1: Did you intend to check in this file?

src/Filtering/nonlinear_diffusion.jl
```diff
@@ -65,6 +65,18 @@ function nonlinear_diffusion(
     return nonlinear_diffusion(img, PeronaMalikDiffusion(λ, K, niters, "inverse_quadratic"))
 end
 
+function nonlinear_diffusion(
+    img::AbstractArray{<:Union{AbstractRGB,TransparentRGB,AbstractGray}},
```
cpaniaguam, line 69: Can we introduce an alias for this long type in this file? Perhaps something like
```julia
const DiffusionPixel = Union{AbstractRGB, TransparentRGB, AbstractGray}
const DiffusionImage = AbstractArray{<:DiffusionPixel}
```

src/Filtering/nonlinear_diffusion.jl
```diff
@@ -65,6 +65,18 @@ function nonlinear_diffusion(
...
+    tiles,
+    f::AbstractDiffusionAlgorithm=PeronaMalikDiffusion()
+)
+    out = deepcopy(img)
```
cpaniaguam, line 73: Because the objects in the array are immutable, `copy` is enough but it doesn't matter anyway.

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
-    IceDetectionBrightnessPeaksMODIS134(band_1_min=0.3, window_size=3, minimum_prominence=0.01, nbins=64)
-    IceDetectionBrightnessPeaksMODIS134(truecolor_image)
-    IceDetectionBrightnessPeaksMODIS134(truecolor_image, tiles)
-
-Identifies sea ice using the midpoint of the `band_1_min` and the detected band 1 mode. Expects the modis truecolor image as input.
-In the case that there are no peaks above the `band_1_min`, uses `band_1_min` as a simple threshold. The `window_size` and `minimum_prominence`
-are passed on to the function [`get_ice_peaks`](@ref). The input image should already have cloud and land masks applied. Optionally, a list of 
-tiles can be provided and the algorithm will be run on each tile.
-
+    IceDetectionBrightnessMidpoint(minimum_reflectance=0.3, window_size=3, minimum_prominence=0.01, nbins=64)
```
cpaniaguam:
```suggestion
    IceDetectionBrightnessMidpoint(minimum_reflectance=0.3, window_size=3, minimum_prominence=0.01, nbins=128)
```

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
+    IceDetectionBrightnessMidpoint(grayscale_image)
+    IceDetectionBrightnessMidpoint(grayscale_image, tiles)
+
+Binarize image using the midpoint of the `minimum_brightness` and the detected mode in the reflectance histogram. Intended for use with MODIS
+Band 1 or Band 2 (e.g., visible red or near-infrared bands). In the case that there are no peaks above the `band_1_min`, uses `band_1_min` as a
+simple threshold. The `window_size` and `minimum_prominence` are passed on to the function [`get_ice_peaks`](@ref). For sea ice binarization,
+input image should already have cloud and land masks applied. Optionally, a list of tiles can be provided and the algorithm will be run on each tile.
+
+```julia
+a = IceDetectionBrightnessPeaks(0.3, 3, 0.01, 128)
```
cpaniaguam:
```suggestion
a = IceDetectionBrightnessMidpoint(0.3, 3, 0.01, 128)
```

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
-Identifies sea ice using the midpoint of the `band_1_min` and the detected band 1 mode. Expects the modis truecolor image as input.
-In the case that there are no peaks above the `band_1_min`, uses `band_1_min` as a simple threshold. The `window_size` and `minimum_prominence`
-are passed on to the function [`get_ice_peaks`](@ref). The input image should already have cloud and land masks applied. Optionally, a list of 
-tiles can be provided and the algorithm will be run on each tile.
-
+    IceDetectionBrightnessMidpoint(minimum_reflectance=0.3, window_size=3, minimum_prominence=0.01, nbins=64)
+    IceDetectionBrightnessMidpoint(grayscale_image)
+    IceDetectionBrightnessMidpoint(grayscale_image, tiles)
+
+Binarize image using the midpoint of the `minimum_brightness` and the detected mode in the reflectance histogram. Intended for use with MODIS
```
cpaniaguam:
```suggestion
Binarize image using the midpoint of the `minimum_reflectance` and the detected mode in the reflectance histogram. Intended for use with MODIS
```

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
+Band 1 or Band 2 (e.g., visible red or near-infrared bands). In the case that there are no peaks above the `band_1_min`, uses `band_1_min` as a
+simple threshold. The `window_size` and `minimum_prominence` are passed on to the function [`get_ice_peaks`](@ref). For sea ice binarization,
```
cpaniaguam: Consider revising this part. `band_1_min` doesn't seem to be an arg of this algo.

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
-    nbins=64
+    nbins=128
 end
 
-function (f::IceDetectionBrightnessPeaksMODIS134)(
-    truecolor_image::AbstractArray{<:Union{AbstractRGB,TransparentRGB}}
-)
-    banddata = red.(truecolor_image)
-    edges, bincounts = build_histogram(banddata, f.nbins; minval=0, maxval=1)
+function (f::IceDetectionBrightnessMidpoint)(out, gray_image::AbstractArray{<:Union{AbstractGray, TransparentGray}})
```
cpaniaguam, line 289: Consider aliases for the long types in this file as well. It seems like the one I suggested above is general enough to be reused here.

src/Segmentation/ice-detection-algorithms.jl
```diff
@@ -264,46 +264,60 @@ function tiled_adaptive_binarization(
...
+    alpha_binary = alpha.(alphacolor.(gray_image)) .> 0.5
+    alpha_binary = 1
+    edges, bincounts = build_histogram(gray_image .* alpha_binary, f.nbins; minval=0, maxval=1)
```
cpaniaguam, line 291: Consider revising this. Is `alpha_binary needed`?

src/Tracking/filter_functions.jl
```diff
@@ -233,43 +233,64 @@ applying 7 individual [`AbstractFloeFilterFunctions`](@ref) in sequence:
 Filters in step 2 use the [`PiecewiseLinearThresholdFunction`](@ref) for thresholds, while Filter 1 uses a [`LinearTimeDistanceFunction`](@ref).
 """ # TODO: Add reference to cal-val paper when ready.
 function FilterFunction()
-    return ChainedFilterFunction(;
+    ChainedFilterFunction(;
         filters=[
-            DistanceThresholdFilter(),
+            DistanceThresholdFilter(
```
cpaniaguam: Could we refactor this filters vector into named pieces? It'd make it easier to read.

test/test-find-ice-labels.jl
```diff
@@ -48,39 +48,39 @@
...
         case = first(filter(c -> (c.case_number == 111 && c.satellite == "terra"), dataset))
         floes = validated_binary_floes(case) .> 0
         clouds = Watkins2025CloudMask()(modis_falsecolor(case))
         land = modis_landmask(case) .> 0
 
-        tc_masked = apply_cloudmask(RGB.(modis_truecolor(case)), clouds)
-        tc_masked .= apply_landmask(tc_masked, land)
-        prelim_ice = f(tc_masked)
+        tc_masked = masker(land, apply_cloudmask(RGB.(modis_truecolor(case)), clouds))
+        # tc_masked .= apply_landmask(tc_masked, land)
```
cpaniaguam: Delete this line?

test/test-find-ice-labels.jl
```diff
@@ -48,39 +48,39 @@
...
+        prelim_ice = f(Gray.(red.(tc_masked))) .> 0
         recall =
             sum(prelim_ice .&& floes .&& .! clouds .&& .! land) / sum(floes .&& .! clouds .&& .! land)
```
cpaniaguam: Consider adding a helper for the recall calculation and reuse it.

test/test-find-ice-labels.jl
```diff
@@ -48,39 +48,39 @@
...
         @test recall >= 0.979
         water = sum(.! prelim_ice .&& .! clouds .&& .! land) ./ prod(size(land))
```
cpaniaguam, line 79: Also for the water calculation.

test/test-find-ice-labels.jl
```diff
@@ -48,39 +48,39 @@
...
         @test 0.38 < water < 0.40
 
         tiles = get_tiles(land, 200)
-        prelim_ice = f(tc_masked, tiles)
+        prelim_ice = f(Gray.(red.(tc_masked)), tiles) .> 0
```
cpaniaguam, line 83:
```suggestion
        prelim_ice = @. f(Gray(red(tc_masked)), tiles) > 0
```

workflow/configs/validation-cases/validation-region-case.csv
```diff
@@ -1,190 +1,190 @@
...
+364,183,sea_of_okhostk,2021-04-05,148.8990631,55.9328092,-912500,3687500,2021-04-05,2021-04-05,250,EPSG:4326,sea_of_okhostk_183,EPSG:3413,FSPipeline,100_000,100_000
+366,184,sea_of_okhostk,2009-05-24,153.4349488,58.32371433,-1112500,3337500,2009-05-24,2009-05-24,250,EPSG:4326,sea_of_okhostk_184,EPSG:3413,FSPipeline,100_000,100_000
+368,185,sea_of_okhostk,2011-05-04,154.4742141,58.58991723,-1162500,3287500,2011-05-04,2011-05-04,250,EPSG:4326,sea_of_okhostk_185,EPSG:3413,FSPipeline,100_000,100_000
+370,186,sea_of_okhostk,2017-05-23,155.1828691,59.601128,-1162500,3162500,2017-05-23,2017-05-23,250,EPSG:4326,sea_of_okhostk_186,EPSG:3413,FSPipeline,100_000,100_000
+372,187,sea_of_okhostk,2013-04-29,155.7309555,59.72775118,-1187500,3137500,2013-04-29,2013-04-29,250,EPSG:4326,sea_of_okhostk_187,EPSG:3413,FSPipeline,100_000,100_000
+374,188,sea_of_okhostk,2015-03-12,156.8383984,57.66080433,-1337500,3337500,2015-03-12,2015-03-12,250,EPSG:4326,sea_of_okhostk_188,EPSG:3413,FSPipeline,100_000,100_000
+376,189,sea_of_okhostk,2012-04-26,163.0858208,61.41718106,-1487500,2787500,2012-04-26,2012-04-26,250,EPSG:4326,sea_of_okhostk_189,EPSG:3413,FSPipeline,100_000,100_000
```
cpaniaguam:
```suggestion
376,189,sea_of_okhostk,2012-04-26,163.0858208,61.41718106,-1487500,2787500,2012-04-26,2012-04-26,250,EPSG:4326,sea_of_okhostk_189,EPSG:3413,FSPipeline,100_000,100_000

```

test/test-segmentation-FSPipeline.jl
```diff
@@ -0,0 +1,130 @@
...
+    @test results_invariant_for(n0f16; baseline, algorithm, case)
+    @test results_invariant_for(float32; baseline, algorithm, case)
+    @test results_invariant_for(float64; baseline, algorithm, case)
+    @test results_invariant_for(RGB, n0f8; baseline, algorithm, case)
+    @test results_invariant_for(RGB, n6f10; baseline, algorithm, case)
+    @test results_invariant_for(RGB, n4f12; baseline, algorithm, case)
+    @test results_invariant_for(RGB, n2f14; baseline, algorithm, case)
+    @test results_invariant_for(RGB, n0f16; baseline, algorithm, case)
+    @test results_invariant_for(RGB, float32; baseline, algorithm, case)
+    @test results_invariant_for(RGB, float64; baseline, algorithm, case)
```
cpaniaguam: Consider parameterizing these.

```suggestion
    paired_types = [n0f8, n6f10, n4f12, n2f14, n0f16, float32, float64]
    unary_types = [RGB, RGBA, paired_types...]
    
    for T in unary_types
        @test results_invariant_for(T; baseline, algorithm, case)
    end
    
    for T in paired_types
        @test results_invariant_for(RGB, T; baseline, algorithm, case)
    end
```

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+            df_rs[:, :s1_perimeter] .= floe.perimeter
+            df_rs[:, :s1_row_centroid] .= floe.row_centroid
+            df_rs[:, :s1_col_centroid] .= floe.col_centroid
+            df_rs[:, :dist_s1_s2] = euclidean_distance(floe, df_rs; r=1) # use pixel units, not meters
+            df_rs[:, :scaled_relative_error_area] = abs.(df_rs.area .- floe.area) ./ (df_rs.area .+ floe.area)
+            push!(results, df_rs)
+        end
+    end
+    
+    results_df = vcat(results..., cols=:union)
```
cpaniaguam: I think we had some bugs with `vcat` on empty collections. Is that something that may happen here?

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+    falsecolor_image = float64.(RGB.(falsecolor))
+    landmask = landmask .> 0 # make sure it's a bitmatrix, in case it's passed as Gray
+    apply_landmask!(truecolor_image, landmask)
+    apply_landmask!(falsecolor_image, landmask)
+
+    n, m = size(truecolor_image)
+    tile_size_pixels = s.tile_size_pixels
+    tile_size_pixels > maximum([n, m]) && begin
+        @warn "Tile size too large, defaulting to image size"
+            tile_size_pixels = minimum([n, m])
```
cpaniaguam:
```suggestion
    nmin, nmax = extrema([n, m])
    tile_size_pixels > nmax && begin
        @warn "Tile size too large; clamping to min(height, width)."
        tile_size_pixels = nmin
```

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+        min_floe_size=100,
+        min_cloudy_floe_size=1000,
+        max_floe_size=50_000,
+        min_band_2_reflectance=0.4,
+        min_cloudy_band_2_reflectance=0.7,
+        cloud_frac_threshold=0.5,
+        min_circularity=0.3,
+        min_cloudy_circularity=0.5
+    )
+end
```
cpaniaguam: Consider hoisting these default algorithm instances/parameter tuples to module-level consts and reference them from Segment defaults. Const defaults would make calibration values easier to audit and (possibly) reuse later.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+    # Re-label so there are no missing numbers in the component list
+    final_floes .= label_components(final_floes)
+
+    # Return the original truecolor image, segmented
+    segments_tc = SegmentedImage(truecolor_image, final_floes)
+    segments_fc = SegmentedImage(falsecolor_image, final_floes)
+
+    if !isnothing(intermediate_results_callback)
+        colorview_truecolor = view_seg(segments_tc)
+        colorview_falsecolor = view_seg(segments_fc)
```
cpaniaguam: These seem to not be used?

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+         "s2_perimeter", 
+         "s2_col_centroid",
+         "s2_row_centroid",
+         "s2_circularity",
+         "s2_mean",
+         "s2_bdry_mean",
+         "s2_bdry_contrast",
+         "dist_s1_s2",
+         "scaled_relative_error_area"
+    ]
```
cpaniaguam: Hoist to const.

src/Pipeline/FSPipeline.jl
cpaniaguam, line 1: Could we split this file into more focused files/modules (for example preprocess, segment, merge, track) and keep `FSPipeline.jl` as a thin composition layer? It would make this much easier to review and maintain.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,859 @@
...
+function (p::Preprocess)(
+    truecolor_image::AbstractArray{<:Union{AbstractRGB,TransparentRGB}}, 
+    landmask,
+    tiles
+)
+    # Cast to grayscale first to save compute time
+    proc_img = Gray.(truecolor_image)
+    
+    # Diffusion and sharpening
+    nonlinear_diffusion(proc_img, tiles, p.diffusion_algorithm)
```
cpaniaguam: Is this supposed to mutate `proc_img`. Per the implementation in `Filtering/nonlinear_diffusion.jl` it seems like it might need to be reassigned.

cpaniaguam (dismissed):

fspipeline_test.jl
danielmwatkins: Ah, thanks! That was meant for checking the functionality during development.

src/Filtering/nonlinear_diffusion.jl
danielmwatkins: Worth thinking about across the board -- I have it written this way to match how other functions in the package are written (e.g., `imbrighten`, `binarize_mask`, `apply_cloudmask!`). I'll add an issue to consider simplifying inputs. I see that John has some functions with `AbstractArray{<:Colorant}` which may be the generalization we're looking for.

src/Pipeline/FSPipeline.jl
danielmwatkins: ah! good catch, that's important

src/Pipeline/FSPipeline.jl
danielmwatkins: Yes and no. In my view, the core components are Preprocess, Segment, and Track. The Preprocess function is as thin as it can get - it's just calling three functions that are defined elsewhere in the code in sequence.  The benefit here is that it makes it possible to change the preprocessing without rewriting the segmentation function, which is not possible in the LopezAcosta workflows. Tracking is also very thin -- it's purely a list of filter functions and settings to call the floe tracker function, no new methods are defined at all. 

For the segmentation section, my thought was to follow a similar development path as I've done with the other Pipelines. There, we first made a working version with a bunch of sub-functions, then when we identified the places where generalization was appropriate, we moved functions into the core package. 

I've defined the following main helper functions which are used in Segment:
* `clean_binary_floes`
* `dist_morph_split`
* `filter_floes!`
* `merge_floes`

There are versions of the "clean binary floes" idea in other places of the code. I see that as a typical "helper function" which is not as generalized and is just defined and used in a Pipeline. 

`dist_morph_split` is the primary one with a direct path to generalization. I've taken the step of giving it a docstring, but have not yet gone as far as coming up with a new type of abstract algorithm for the floe splitting task. Time constraints mean that I don't think that I can do sufficient generalization in the short term, but in the medium term it should be doable to move it into one of the core modules, likely in the Segmentation module. There's an analogous function in LopezAcosta2019, where I was able to identify a standalone section in the original code. These two could be the basis for a FloeSplittingAlgorithm section. Similarly, later on, we could make a Julia version of the `ebseg` erosion-expansion method, since that is also a floe splitting algorithm.

The `filter_floes!` function also could maybe be generalized; neither of the other algorithms take advantage of object-based analysis. It would take some creativity, thought, and time to figure out the right level of generalization here.

Finally, the `merge_floes` is a core idea which could be generalized. It's got sub-functions which are especially important and have potential for being generally-useful tools once we have time to think through the different aspects and approaches for each.
* `assign_labels` -- short and simple, could be useful elsewhere potentially. 
* `get_relevant_set` -- this could be a segmentation utility, or go in a metrics module. This should eventually be a core function, though.
* `objectwise_compare_segmentation` -- there is substantial overlap in concept with the Tracking functions, with making candidate floe dataframes. With enough time, we may be able to figure out a uniform approach to making comparisons between sets. The best way to do that is probably to think critically about both the objectwise comparison and the tracker function. And defining more functions that add the comparison metrics. This will be generally useful.

src/Segmentation/ice-detection-algorithms.jl
danielmwatkins: I added an issue, this is something that can be considered throughout the code base.

src/Tracking/filter_functions.jl
danielmwatkins: No, the reason being that the point of the ChainedFilterFunction is that it takes whatever filters you want to apply and then does them in order. Adding named pieces would mean that you no longer have the flexibility to change the order of filters.

test/test-find-ice-labels.jl
danielmwatkins: I simplified the calculation a bit, adding a recall function in a segmentation metrics section later would be a good idea.

src/Segmentation/ice-detection-algorithms.jl
danielmwatkins: Oh right, this was an issue with trying to figure out the `masker` alpha channel methods.

test/test-find-ice-labels.jl
danielmwatkins: Simplified the equation, as with the recall function. I think with these two cases it's not really needed to have separate named functions.

test/test-segmentation-FSPipeline.jl
danielmwatkins: This is all John's code, but I love the shortening with the for loops. Good call. The main benefit / reason for doing the single version at a time is that in other pipelines, some are broken. But here it all works so we're fine.

src/Tracking/filter_functions.jl
cpaniaguam: I put a concrete version of my vision in https://github.com/WilhelmusLab/IceFloeTracker.jl/pull/1026. As you can see, one can always change the order of things in the definition of the `filters` vector.

src/Pipeline/FSPipeline.jl
danielmwatkins: Good catch, I added a step to return an empty dataframe with the right columns in the case.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,952 @@
...
+            PsiSCorrelationThresholdFilter(;
+                threshold_function=PiecewiseLinearThresholdFunction(;
+                    minimum_area=100,
+                    maximum_area=700,
+                    minimum_value=0.86,
+                    maximum_value=0.96,
+                ),
+            ),
+        ],
+    ),
```
cpaniaguam: Is this the same filter used in `src/Tracking/filter_functions.jl`? If so, it should be reused here.

cpaniaguam (changes requested):

src/Pipeline/FSPipeline.jl
danielmwatkins: Yes, and no. I'm placing it here because I'm going to be changing the defaults after a few tests. So it currently matches, but won't always. Setting new defaults requires writing out just as much code, since each filter function has to get written out and initialized. I want all the settings to be easily seen here.

cpaniaguam: Okay. Consider the pattern to build it piecemeal in as in `src/Tracking/filter_functions.jl`. I think it'd be useful to revise it later.

danielmwatkins: That's a good idea. I've added that, it's a better format. Thanks for the suggestion.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+                end
+            else
+                if b2_means[L] < min_band_2_reflectance
+                    img_indexmap[indices[L]] .= 0
+                elseif circ[L] < min_circularity
+                    img_indexmap[indices[L]] .= 0
+                end
+            end
+        end
+    end
```
cpaniaguam, line 516: Less indented.
```suggestion
    for L in unique(img_indexmap)
        L <= 0 && continue
        if cloud_fractions[L] > cloud_frac_threshold
            if areas[L] < min_cloudy_floe_size
                img_indexmap[indices[L]] .= 0
            elseif b2_means[L] < min_cloudy_band_2_reflectance
                img_indexmap[indices[L]] .= 0
            elseif circ[L] < min_cloudy_circularity
                img_indexmap[indices[L]] .= 0
            end
            continue
        end

        if b2_means[L] < min_band_2_reflectance
            img_indexmap[indices[L]] .= 0
        elseif circ[L] < min_circularity
            img_indexmap[indices[L]] .= 0
        end
    end
```

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+                if (0 ∈ matched_labels) && (length(matched_labels) <= 2)
+                    final_labels[indices[L]] .= L
+                else
+                    # Otherwise, expand the current level, and set the next level down to the expanded indices.
+                    # May need to check the number of matched labels in the expanded image.
+                    levels[dist_threshold - 1][indices[L]] .= expanded[indices[L]]
+                    final_labels[indices[L]] .= expanded[indices[L]]
+                end
+            end
+        end
```
cpaniaguam, line 445:
```suggestion
        for L in keys(indices)
            (L <= 0) && continue

            matched_labels = unique(levels[dist_threshold][indices[L]])

            # If intersection of the label at level
            if (0 ∈ matched_labels) && (length(matched_labels) <= 2)
                final_labels[indices[L]] .= L
                continue
            end
            # Otherwise, expand the current level, and set the next level down to the expanded indices.
            # May need to check the number of matched labels in the expanded image.
            levels[dist_threshold - 1][indices[L]] .= expanded[indices[L]]
            final_labels[indices[L]] .= expanded[indices[L]]
        end
```

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+                cmin = minimum((floe.min_col, s_floe.min_col))
+                cmax = maximum((floe.max_col, s_floe.max_col))
+
+                # check if area overlap between g and s is larger than 50% of g
+                gtmask = labels1[rmin:rmax, cmin:cmax] .== floe.label
+                slmask = labels2[rmin:rmax, cmin:cmax] .== s_floe.label
+                intersect_area = sum(gtmask .&& slmask)
+                maximum([intersect_area / s_floe.area, intersect_area / floe.area]) > 0.5 && begin
+                    push!(relevant_set_labels, s_floe.label)
+                end
```
cpaniaguam:
```suggestion
                if maximum([intersect_area / s_floe.area, intersect_area / floe.area]) > 0.5
                    push!(relevant_set_labels, s_floe.label)
                end
```

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+function get_relevant_set(df1, df2, labels1, labels2)
+    relevant_set = Dict{Int64,Vector{Int64}}()
+    for floe in eachrow(df1)
+        # select labels that are inside the bounding box for the floe
+        matched_labels = unique(
+            labels2[floe.min_row:floe.max_row, floe.min_col:floe.max_col]
+        )
+
+        # if any, then check centroid positions
+        maximum(matched_labels) != 0 && begin
```
cpaniaguam: Here the guard clause pattern could be used and avoid indenting the block further.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+        "s2_perimeter",
+        "s2_col_centroid",
+        "s2_row_centroid",
+        "s2_circularity",
+        "s2_mean",
+        "s2_bdry_mean",
+        "s2_bdry_contrast",
+        "dist_s1_s2",
+        "scaled_relative_error_area",
+    ],
```
cpaniaguam: I would move the list of names to a const outside of the function just to clean up the functions signature.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+        results_df,
+        :area => :s2_area,
+        :perimeter => :s2_perimeter,
+        :label => :s2_label,
+        :col_centroid => :s2_col_centroid,
+        :row_centroid => :s2_row_centroid,
+        :max_col => :s2_max_col,
+        :max_row => :s2_max_row,
+        :min_col=>:s2_min_col,
+        :min_row=>:s2_min_row,
```
cpaniaguam: I would also move this mapping out to a const and build it programmatically (add s_2 prefixes).

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+            :scaled_distance,
+            :relative_error_area,
+            :relative_error_convex_area,
+            :relative_error_major_axis_length,
+            :relative_error_minor_axis_length,
+            :psi_s_correlation_score,
+            :scaled_shape_difference,
+        ],
+        weights=ones(7),
+    ),
```
cpaniaguam, line 918: Consider defining the matching function outside of the functions signature for a cleaner definition.

cpaniaguam (changes requested):

src/Pipeline/FSPipeline.jl
danielmwatkins: I always get mixed up with the scope with Julia if statements and `continue`, but I do like how this looks with your changes.

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+
+        # Clear intersections
+        idx = F .> 0
+        for L in filter(r -> r != 0, unique(A[idx]))
+            A[A_indices[L]] .= 0
+        end
+        for L in filter(r -> r != 0, unique(B[idx]))
+            B[B_indices[L]] .= 0
+        end
+    end
```
cpaniaguam, line 826: I think this block is hard to parse by a person. I'll create an issue to clean this up later.

cpaniaguam: #1031

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,960 @@
...
+
+"""
+function merge_floes(indexmap1, indexmap2, img; dmax=10, emax=0.25, min_floe_size=100)
+
+    # If no floes to merge, skip merge
+    if maximum(indexmap1) == 0
+        return indexmap2
+    elseif maximum(indexmap2) == 0
+        return indexmap1
+    end
```
cpaniaguam:
```suggestion
    maximum(indexmap1) == 0 && return indexmap2
    maximum(indexmap2) == 0 && return indexmap1
```

src/Pipeline/FSPipeline.jl
danielmwatkins: Redesigned to use a loop and concatenation of Symbols

src/Pipeline/FSPipeline.jl
```diff
@@ -0,0 +1,928 @@
...
+function get_relevant_set(df1, df2, labels1, labels2)
+    relevant_set = Dict{Int64,Vector{Int64}}()
+    for floe in eachrow(df1)
+        # select labels that are inside the bounding box for the floe
+        matched_labels = unique(
+            labels2[floe.min_row:floe.max_row, floe.min_col:floe.max_col]
+        )
+
+        # if any, then check centroid positions
+        maximum(matched_labels) != 0 && continue
```
cpaniaguam: Sorry, this one I think needs to be inverted to skip early:

```suggestion
        maximum(matched_labels) == 0 && continue
```

danielmwatkins: Ha, I'd caught that immediately and forget to push the update. Glad you caught it as well.

cpaniaguam (approved):
Thanks!

PR #971, opened 2026-05-20
fix: regionprops table output should include all implied columns even if there are no rows

hollandjg:
Output all implied columns including `row_centroid`, as required to fix the error in #956.

### DIFF
--- a/test/test-regionprops.jl
+++ b/test/test-regionprops.jl
@@ -90,7 +90,6 @@ end
 
     @testset "bbox" begin
         table = regionprops_table(label_img; properties=["bbox"])
-        @test nrow(table) == 0
         @test "min_row" ∈ names(table)
         @test "max_row" ∈ names(table)
         @test "min_col" ∈ names(table)
--- a/test/test-regionprops.jl
+++ b/test/test-regionprops.jl
@@ -70,7 +70,7 @@ end
 @testitem "regionprops table output should include all implied columns even if there are no rows" begin
     using DataFrames: nrow
 
-    label_img = Int.(zeros(5, 5))
+    label_img = zeros(Int, 5, 5)
     @testset "defaults" begin
         table = regionprops_table(label_img)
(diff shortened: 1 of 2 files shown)


test/test-regionprops.jl
```diff
@@ -67,6 +67,43 @@
     )
 end
 
+@testitem "regionprops table output should include all implied columns even if there are no rows" begin
+    using DataFrames: nrow
+
+    label_img = Int.(zeros(5, 5))
```
cpaniaguam:
```suggestion
    label_img = zeros(Int, 5, 5)
```

test/test-regionprops.jl
```diff
@@ -67,6 +67,43 @@
...
+        @test "major_axis_length" ∈ names(table)
+        @test "minor_axis_length" ∈ names(table)
+        @test "convex_area" ∈ names(table)
+        @test "perimeter" ∈ names(table)
+        @test "orientation" ∈ names(table)
+    end
+
+    @testset "bbox" begin
+        table = regionprops_table(label_img; properties=["bbox"])
+        @test nrow(table) == 0
```
cpaniaguam: Should this be the case only when "bbox" is included?

cpaniaguam (approved):
lgtm

test/test-regionprops.jl
hollandjg: The table will have zero rows if the labels are empty. It should definitely have all the `{min,max}_{row,col}` names if `bbox` is included. 

If neither `bbox` nor `{min,max}_{row,col}` are in the properties list, then they shouldn't be in the output dataframe.

cpaniaguam: Thanks for the clarification. But I meant the check for 0 rows. I was curious why check that condition in this test case only.

hollandjg: Oh true!

PR #972, opened 2026-05-20
refactor: reorganize snakemake results by region scene filetype

hollandjg:
Update the results directory to include a higher level of directory organization:
Screenshot 2026-05-20 at 4 48 32 PM

Files are organized by:
- `region` for landmasks which don't change from scene to scene, or batches of overpass times
- `scene` for files which change each observation timepoint
- `track` for tracked results from multiple scenes
- `filetype` for hardlinked copies of the files from `scene`, subdivided into 
  - different image types, like `landmask`, `truecolor` etc., and 
  - pipelines and their outputs, like `LopezAcosta2019/labels_map`
- `init` for files generated in advance of any computation while initializing the workflow environment


workflow/snakefile
```diff
@@ -132,31 +100,64 @@ rule region_case_results_by_filetype:
...
+    output: 
+        "filetype/{filetype}/{scene}.{ext}"
+    input:
+        "scene/{scene}/{filetype}.{ext}"
+    wildcard_constraints:
+        scene = "[^/]+",
+        filetype = "[^./]+",
+        pipeline = "[^/]+"
+    shell:
+        "ln --force {input} {output}"
```
danielmwatkins: Does this need to be -f instead?

hollandjg:
```suggestion
        "ln -f {input} {output}"
```

hollandjg: Yes! Good catch.

danielmwatkins (approved):
Looks good!

PR #973, opened 2026-05-20
feat: add HDF5 loading

hollandjg:
Add support for re-loading HDF5 files from disk.

Also needed to add a `binarize_mask` which works on Integer arrays


src/PersistHDF5/PersistHDF5.jl
```diff
@@ -29,6 +30,7 @@ end
     landmask::AbstractMatrix
     coastal_buffer_mask::AbstractMatrix
     iftversion::VersionNumber = pkgversion(@__MODULE__)
+    file_version::VersionNumber = VersionNumber("1.0.0")
```
cpaniaguam, line 33: Can you help me understand what `file_version` is for?

test/test-make_hdf5.jl
```diff
@@ -50,6 +50,35 @@
...
+        @test reloaded.cloud_mask == data.cloud_mask
+        @test reloaded.ice_mask == data.ice_mask
+        @test reloaded.landmask == data.landmask
+        @test reloaded.coastal_buffer_mask == data.coastal_buffer_mask
+        @test reloaded.iftversion == data.iftversion
+        @test reloaded.reference == data.reference
+        @test reloaded.contact == data.contact
+
+        @show reloaded.landmask.size, data.landmask.size
+        @show eltype(reloaded.landmask), eltype(data.landmask)
```
cpaniaguam: Should these be tests?

test/test-make_hdf5.jl
```diff
@@ -50,6 +50,35 @@
             @test haskey(file["classifications"], "ice_mask")
             @test haskey(file["classifications"], "coastal_buffer_mask")
         end
```
cpaniaguam: Consider removing these checks as they are included when `reloaded` is built via `load_hdf5` below.

cpaniaguam (dismissed):
Couple of comments, thanks!

src/PersistHDF5/PersistHDF5.jl
hollandjg: This is so that when the file format we want changes, the file parser can read the version number and can load exactly the right things, without having to introspect on what's in the file, or have a lot of complicated error handling logic for when things are missing. 

Format changes might include:
- adding new images or tables
- renaming attributes
- changing the data types we use for images or tables.

This use case isn't covered by the package version number because the package will change at a different rate. There might be a "new" version of IceFloeTracker.jl with a key bugfix, and a new file format, but someone just wants to rewrite some existing HDF5 files with the bugfix, but without using the newer file format. We can support that if we have different versions of the file format which are independent of the package version.

test/test-make_hdf5.jl
hollandjg: They can go away.

test/test-make_hdf5.jl
hollandjg: These are meant to serve a different purpose. They check the HDF5 format explicitly, i.e. that there are particular groups with particular names and nested keys. That part is `file_version`-specific. 

The tests below check that the `load_hdf5` function returns equivalents of the objects which went in, without caring about the structure of the file. That part is `file_version`-independent.

hollandjg: I've added testsets to wrap those two blocks and label what they are for.

hollandjg: I've added a file-version check to the top block.

danielmwatkins (approved):

PR #974, opened 2026-05-20
refactor: make HDF5 saver work on a struct

hollandjg:
Update the call for the `make_hdf5` file to potentially support different HDF5 structures.


cpaniaguam (approved):
I like this abstraction to simplify the functions signature, thanks!

danielmwatkins (approved):
I like this as well.

PR #975, opened 2026-05-20
refactor: rename HDF5 from PersistHDF5

hollandjg:
Make the "CLI" use of the HDF5 submodule a bit clearer by renaming the submodule to just HDF5. It shouldn't collide at all with the real `HDF5` module


cpaniaguam (approved):

danielmwatkins (approved):
Sounds good!

PR #976, opened 2026-05-20
refactor: rename `save_hdf5` function

hollandjg:
Used to be `make_hdf5`, but `save_hdf5` is closer to the FileIO standard we're trying to keep close to. It's not straightforward to make this an overload of `FileIO.save`, but we may come back to that.


src/ImageUtils/mask.jl
```diff
@@ -219,3 +219,20 @@ function binarize_mask(
...
+Assumes that the input image is 0 over the parts to be left unmasked, and some shade over the parts to be masked. 
+The tol argument lets a higher threshold for masked pixels be chosen.
+
+# Arguments
+- `mask_image`: mask image, e.g. loaded from an 8-bit TIFF
+- `tol` (Optional): Values in the image larger than `tol` are set to true (1).
+"""
+function binarize_mask(mask_image::AbstractArray{<:Integer}; tol=0)::BitMatrix
+    return mask_image .> tol
+end
```
cpaniaguam:
```suggestion
end

```

cpaniaguam (approved):

PR #985, opened 2026-06-02
refactor[workflow]: add an independent julia project in the workflow results directory

hollandjg:
A new Julia environment is built in the working directory of the Snakemake workflow.


workflow/snakefile
```diff
@@ -159,7 +159,21 @@ rule init_ift:
...
+        using Pkg; 
+        {config[IFT_INSTALL]}
+        Pkg.add([
+            "CSVFiles",
+            "DataFrames",
+            "HDF5",
+            "Images",
+            "TimeZones",
+            "FileIO",
+        ])
```
cpaniaguam: Could these workflow dependencies live in a dedicated Project.toml instead of being added inline here? It would make the workflow environment more reproducible than resolving packages ad hoc during each run.

hollandjg: I agree it would be great. But it doesn't work, because of the way Julia handles its Project and Manifest files. We need to be able to specify a path to an IceFloeTracker.jl repo, or a revision on GitHub, or a version in the Julia registry. 
The only way I know to do that cleanly is by using the install commands `Pkg.add` and `Pkg.develop`. 
There isn't a good way I'm aware of to specify that in the Project.toml file, and having it in the Manifest file isn't portable.

I'm happy to revisit this later because I agree this isn't very neat.

hollandjg: How about
```
[deps]
Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
IJulia = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
IceFloeTracker = "04643c7a-9ac6-48c5-822f-2704f9e70bd3"
Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"

[sources]
IceFloeTracker = { path = ".." }

[compat]
Documenter = "1"
```

hollandjg: - [x] Try this out

hollandjg: Added a `Project.toml` and a mechanism to override it.

.github/workflows/Workflow.yml
```diff
@@ -8,6 +8,10 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+env:
+  SNAKEMAKE: "snakemake --cores=2 --configfile=workflow/configs/ift-from-this-repo/config.yaml"
+
```
cpaniaguam, line 14: Nice abstraction!

cpaniaguam (approved):
Lgtm, thanks!

PR #988, opened 2026-06-03
feat[workflows]: alternative configurations

hollandjg:
Adds new configuration examples:
- default configuration
- An example with both the region definitions and the timepoints in a single CSV file, can be called with `snakemake --configfile workflow/configs/case-region-file/config.yaml`
- [validation dataset](https://github.com/danielmwatkins/ice-floe-validation-dataset/blob/main/data/validation_dataset/validation_dataset.csv) can be called with `snakemake --configfile workflow/configs/validation-cases/config.yaml`.

When calling `snakemake --configfile some/config.yaml`
snakemake will _combine_ `some/config.yaml` with the default configuration. 
So the variables like `IFT` and `LOADER` specified in the default configuration don't need to be duplicated.


workflow/README.md
```diff
@@ -73,6 +73,32 @@ To keep going if a single function in the pipeline fails,
...
+
+It specifies:
+- the [default case list](./configs/default/case.csv)
+- the [default region list](./configs/default/region.csv)
+- which Julia environment is used to run IceFloeTracker.jl commands
+- and other commands used in the workflow.
+
+If an additional configuration is specified on the command line using `--configfile`
+it will be combined with the default configuration
+and can override variables from the default configuration.
```
cpaniaguam, line 88: How are additional config files combined with the default config? Is it simply overriding matching keys from the default, or is there other merge behavior for nested values/lists? It may help to state that explicitly here.

cpaniaguam (commented):
A question about the how the configs interact or override defaults. Thanks!

workflow/README.md
hollandjg: It overrides matching keys from the default. There's no special behavior. All the Snakemake documentation says is: 

> Note that any values parsed into the config dictionary with any of above mechanisms are merged, i.e., all keys defined via a configfile statement, or the --configfile and --config command line arguments will end up in the final config dictionary, but if two methods define the same key, command line overwrites the configfile statement.

danielmwatkins (approved):
Looks good to me

PR #995, opened 2026-06-22
fix[FloeTracker]: Not enough points to compute convex hull.

hollandjg:
Handle convex hull errors correctly. We don't try to catch cases which might throw a convex hull error, because we don't a priori know what those cases will be. Instead, we let the error be thrown, and then we handle it. 

- resolves #919


src/Segmentation/regionprops.jl
```diff
@@ -303,7 +303,13 @@ function (f::PolygonConvexArea)(A)
...
         end
 
-        chull = convexhull(A .== i)
+        local chull
+        try
+            chull = convexhull(A .== i)
+        catch
+            convex_areas[i] = NaN
+            continue
+        end
```
cpaniaguam: Can this workflow be moved to a helper? That way it could be used in both algos. It could look something like
```julia
function _component_convex_hull(A, label)
    try
        return convexhull(A .== label)
    catch err
        if err isa ErrorException &&
           sprint(showerror, err) == "Not enough points to compute convex hull."
            return nothing
        end
        rethrow()
    end
end
```
Note I'd make the error specific to avoid masking other potential unforrseen bugs with `convexhull`. Then both algos could use:
```julia
        chull = _component_convex_hull(A, i)
        isnothing(chull) && begin
            convex_areas[i] = NaN
            continue
        end
```

cpaniaguam (changes requested):

src/Segmentation/regionprops.jl
hollandjg: This is a great idea! I'm going to make it so that the helper function has the same signature as the `convexhull` function and takes a binary array.

cpaniaguam (approved):

PR #997, opened 2026-06-22
fix: make segment mean a fixed-point 8 bit image, rather than float32

hollandjg:
Fix an issue where the segment mean images output in the workflow are of type float32 rather than fixed-point 8-bit

### DIFF
--- a/src/Pipeline/LopezAcosta2019.jl
+++ b/src/Pipeline/LopezAcosta2019.jl
@@ -296,13 +296,14 @@ function (p::Segment)(
         segments_falsecolor = SegmentedImage(falsecolor, labels)
         segment_mean_truecolor=map(i -> n0f8(segment_mean(segments_truecolor, i)), labels)
         segment_mean_falsecolor=map(i -> n0f8(segment_mean(segments_falsecolor, i)), labels)
+        ice_mask=p.cluster_selection_algorithm(fc_masked) .> 0
         intermediate_results_callback(;
             truecolor,
             falsecolor,
             landmask,
             coastal_buffer_mask,
             cloud_mask=cloudmask,
-            ice_mask=p.cluster_selection_algorithm(fc_masked) .> 0,
+            ice_mask,
             sharpened_grayscale_image=sharpened_grayscale_image,
             ice_water_discrim=ice_water_discrim,
             segA=segmentation_A,
--- a/src/Pipeline/LopezAcosta2019.jl
+++ b/src/Pipeline/LopezAcosta2019.jl
@@ -72,7 +72,8 @@ import ..Segmentation:
     kmeans_binarization,
     IceDetectionFirstNonZeroAlgorithm,
     IceDetectionBrightnessPeaksMODIS721,
-    IceDetectionThresholdMODIS721
+    IceDetectionThresholdMODIS721,
+    segment_mean_map
 
 import ..Tracking: FloeTracker, FilterFunction, MinimumWeightMatchingFunction
 import Dates: Day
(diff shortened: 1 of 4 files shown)


src/Pipeline/LopezAcosta2019.jl
```diff
@@ -313,11 +314,12 @@ function (p::Segment)(
...
             segmented_falsecolor=segments_falsecolor,
             segment_mean_truecolor=map( # TODO Add "view_seg" code snippet
-                i -> segment_mean(segments_truecolor, i),
+                i -> n0f8(segment_mean(segments_truecolor, i)),
                 labels_map(segments_truecolor),
             ),
             segment_mean_falsecolor=map(
-                i -> segment_mean(segments_falsecolor, i), labels_map(segments_falsecolor)
+                i -> n0f8(segment_mean(segments_falsecolor, i)),
+                labels_map(segments_falsecolor),
```
cpaniaguam: Why not compute these at the top just like `segments_truecolor/falsecolor`? I would throw in `ice_mask` as well.

src/Pipeline/LopezAcosta2019Tiling.jl
```diff
@@ -307,8 +308,12 @@ function (p::Segment)(
...
             segments_falsecolor,
-            segment_mean_truecolor=map(i -> segment_mean(segments_truecolor, i), labels),
-            segment_mean_falsecolor=map(i -> segment_mean(segments_falsecolor, i), labels),
+            segment_mean_truecolor=map(
+                i -> n0f8(segment_mean(segments_truecolor, i)), labels
+            ),
+            segment_mean_falsecolor=map(
+                i -> n0f8(segment_mean(segments_falsecolor, i)), labels
+            ),
         )
```
cpaniaguam: Hoist to the top?

cpaniaguam (changes requested):

src/Pipeline/LopezAcosta2019.jl
hollandjg: Sure!

src/Pipeline/LopezAcosta2019Tiling.jl
hollandjg: Yes!

src/Pipeline/LopezAcosta2019.jl
```diff
@@ -289,17 +290,21 @@ function (p::Segment)(
...
-
     if !isnothing(intermediate_results_callback)
-        segments_truecolor = SegmentedImage(truecolor, labels)
-        segments_falsecolor = SegmentedImage(falsecolor, labels)
+        segmented_truecolor = SegmentedImage(truecolor, labels)
+        segmented_falsecolor = SegmentedImage(falsecolor, labels)
+        segment_mean_truecolor=map(i -> n0f8(segment_mean(segmented_truecolor, i)), labels)
+        segment_mean_falsecolor=map(
+            i -> n0f8(segment_mean(segmented_falsecolor, i)), labels
+        )
```
cpaniaguam: This mapping operation is applied four times, two times here and another two times in `src/Pipeline/LopezAcosta2019Tiling.jl`. Consider moving it to a helper.

hollandjg: done!

src/Segmentation/segmented-image-utilities.jl
```diff
@@ -247,3 +247,11 @@ function expand_labels(labeled_img::Matrix{Int64}, distance::Int64)
     return labels_out
 end
 
+"""
+    segment_mean_map(s::SegmentedImage, label::Int)
+
+Map each segment in the segmented image `s` to the mean intensity or color of that segment.
+"""
```
cpaniaguam, line 254: This docstring looks stale: the signature still includes `label::Int`, but the method only takes `s::SegmentedImage`. It may also help to mention the returned object more explicitly, either in the signature or in the docstring text.

cpaniaguam (changes requested):
Thanks for the changes! Just one last thing here.

src/Segmentation/segmented-image-utilities.jl
hollandjg: Thanks, good catch.

cpaniaguam (approved):

PR #999, opened 2026-06-24
fix[tracker]: where first dataframe is empty, floes in the first non-empty dataframe get tracked multiple times.

hollandjg:
We have some logic to skip the first tracking dataframe if it's empty. But we don't properly start iterating through the rest of the dataframes, and instead always start at the second. This fixes that error.

### DIFF
--- a/test/test-floe-tracker.jl
+++ b/test/test-floe-tracker.jl
@@ -197,6 +197,84 @@ end
     @test sum(counts[:, :count] .== 3) == 4 && sum(counts[:, :count] .== 2) == 1
 end
 
+@testitem "One blank image at the start of the series" setup = [FloeTrackerBasicCases] begin
+    tracker = FloeTracker(;
+        filter_function=FilterFunction(),
+        matching_function=MinimumWeightMatchingFunction(),
+        minimum_area=floe_area_threshold,
+    )
+
+    # Add full image gap
+    labeled_imgs_gaps = [
+        zeros(Int64, size(labeled_imgs[1])),
+        labeled_imgs[1],
+        labeled_imgs[2],
+        labeled_imgs[3],
+    ]
+
+    # Extend passtimes
+    passtimes_gaps = [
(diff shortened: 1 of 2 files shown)


test/test-floe-tracker.jl
```diff
@@ -197,6 +197,95 @@ end
...
+    labeled_imgs_gaps = [zeros(Int64, size(labeled_imgs[1])), labeled_imgs[1]]
+
+    # Extend passtimes
+    passtimes_gaps = [DateTime("2000-01-01"), DateTime("2000-01-02")]
+
+    tracker = FloeTracker(;
+        filter_function=FilterFunction(),
+        matching_function=MinimumWeightMatchingFunction(),
+        minimum_area=floe_area_threshold,
+    )
```
cpaniaguam: Delete duplicate `tracker` definition.

```suggestion
```

test/test-floe-tracker.jl
```diff
@@ -197,6 +197,95 @@ end
...
+    counts = combine(groupby(trajectories, [:ID]), nrow => :count)
+    @test nrow(trajectories) == 0
+end
+
+@testitem "Several blank images at the start of the series" setup = [FloeTrackerBasicCases] begin
+    tracker = FloeTracker(;
+        filter_function=FilterFunction(),
+        matching_function=MinimumWeightMatchingFunction(),
+        minimum_area=floe_area_threshold,
+    )
```
cpaniaguam: It seems like these individual `tracker` definitions could be moved to the `FloeTrackerBasicCases` snippet.

cpaniaguam (changes requested):
Thanks for this! Only two things to consider.

test/test-floe-tracker.jl
hollandjg: That is a fine idea.

cpaniaguam (approved):

PR #1000, opened 2026-06-24
fix[hdf5]: handle cases where dataframe values are `NaN`s

hollandjg:
loading a CSV file using CSVFiles and FileIO (as we do) will replace Float `NaN` with `missing`.
`missing` is incompatible with HDF5 float representations, so before saving to HDF5 we need to 
- convert any `missing` in float columns to `NaN`
- disallow `missing` – make the column a pure Float.


test/test-make_hdf5.jl
cpaniaguam, line 1: Thanks for the integration tests. Consider also adding unit tests for `convert_missing_to_nan` as well.

cpaniaguam (changes requested):
I am approving but I'm going to mark it as "requesting changes" in case automerge is enabled so that a comment doesn't go unread. Thanks!

test/test-make_hdf5.jl
hollandjg: Yes! Let's do that.

cpaniaguam (approved):
Thanks!

PR #1001, opened 2026-06-24
Add try/catch to peak finder

### LINKED ISSUE #952: Error in get ice peaks if the number of histogram bins is too low
Ran into a puzzling error when running through a list of cases. Case 174 aqua is fully cloud covered, and the cloud mask doesn't catch everything. That by itself is fine.

[174-sea_of_okhostk-100km-20220524.aqua.truecolor.250m.tiff](https://github.com/user-attachments/files/27251784/174-sea_of_okhostk-100km-20220524.aqua.truecolor.250m.tiff)

Cloud masked FC image:

Image

When I tried to use the bright ice detector
`IceDetectionBrightnessPeaksMODIS721(band_7_max=0.1, possible_ice_threshold=0.3, join_method="union")`
it errorred. I tracked down the issue to the get_peaks function. 

Image

The issue is that I've set the default number of bins to 64, and the peakproms and peakwidths functions run into issues when the peak is too near a boundary. The issue can be avoided if a smaller window is used (default is window=3) or a larger number of bins is used in the histogram.

danielmwatkins:
Simple try/catch to avoid bug where Peaks can't find the width of peaks in cases such as 174/aqua. 
- resolves #952

### DIFF
--- a/src/Segmentation/find-ice-labels.jl
+++ b/src/Segmentation/find-ice-labels.jl
@@ -58,12 +58,18 @@ function get_ice_peaks(
     # we don't want to include in the normalization. If no potential
     # ice pixels, then return early
     counts = normalizer > 0 ? counts ./ normalizer : return Inf
-    pks = findmaxima(counts, window_size) |> peakproms! |> peakwidths!
+    local pks
+    try
+        pks = findmaxima(counts, window_size) |> peakproms! |> peakwidths!
+    catch
+        return Inf
+    end
     pks_df = DataFrame(pks[Not(:data)])
     pks_df = sort(pks_df, :proms; rev=true)
     mx, argmx = findmax(pks_df.proms)
     mx < minimum_prominence && return Inf
     return edges[pks_df[argmx, :indices]]
+
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/Segmentation/find-ice-labels.jl
```diff
@@ -58,12 +58,18 @@ function get_ice_peaks(
     # we don't want to include in the normalization. If no potential
     # ice pixels, then return early
     counts = normalizer > 0 ? counts ./ normalizer : return Inf
-    pks = findmaxima(counts, window_size) |> peakproms! |> peakwidths!
+    local pks
+    try
+        pks = findmaxima(counts, window_size) |> peakproms! |> peakwidths!
+    catch
+        return Inf
+    end
```
cpaniaguam, line 68: This bare `catch` will capture all other exceptions besides the intended `BoundsError`. Consider making it specific.

```suggestion
    local pks
    try
        pks = findmaxima(counts, window_size) |> peakproms! |> peakwidths!
    catch e
        e is a BoundsError || rethrow()
        @debug "Peak finder failed (peak near boundary), returning `Inf`."
        return Inf
    end
```

hollandjg (approved):
Great!

PR #1002, opened 2026-06-24
fix[workflow]: include tracker results in default rule

hollandjg:
Tracker results had previously been included only when explicitly requested. This includes those results when calling the default snakemake rule, i.e., when calling snakemake without any specific target.


cpaniaguam (approved):
lgtm

PR #1003, opened 2026-06-24
refactor: split tracker test cases into separate testitems

hollandjg:
Just a refactor to make the testcases easier to run independently


cpaniaguam (approved):
👍

PR #1004, opened 2026-06-24
test: add long tracking example cases

hollandjg:
Include an example workflow configuration with long tracking sequences. 

The intention is to allow the user to easily test their version of the pipeline including tracking on a long sequence with a mixture of clear and cloudy observations, ice and no ice.


danielmwatkins (approved):
Looks good to me! Thanks for putting this together.

PR #1007, opened 2026-06-29
fix[tracker]: don't crash rotation measurement if tracker results are empty

hollandjg:
Add handling when the tracker dataframe is empty and we get rotation measurements: return a dataframe of the right shape with no entries.

### DIFF
--- a/src/Tracking/rotation.jl
+++ b/src/Tracking/rotation.jl
@@ -33,16 +33,15 @@ function get_rotation_measurements(
     if nrow(df) == 0
         @debug "No observations in the input DataFrame, returning an empty dataframe with the correct column names"
         original_names, original_types = names(df), eltype.(eachcol(df))
-        new_column_names_types = [
-            ("theta_rad", Float64),
-            ("dt_sec", Float64),
-            ("omega_rad_per_sec", Float64),
-            ("omega_rad_per_day", Float64),
-            ("theta_deg", Float64),
-            ("omega_deg_per_day", Float64),
+        new_column_names = [
+            "theta_rad",
+            "dt_sec",
+            "omega_rad_per_sec",
+            "omega_rad_per_day",
+            "theta_deg",
+            "omega_deg_per_day",
         ]
-        new_column_names = [n for (n, _) in new_column_names_types]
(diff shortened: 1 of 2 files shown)


src/Tracking/rotation.jl
```diff
@@ -30,6 +30,29 @@ function get_rotation_measurements(
...
+        new_column_names_types = [
+            ("theta_rad", Float64),
+            ("dt_sec", Float64),
+            ("omega_rad_per_sec", Float64),
+            ("omega_rad_per_day", Float64),
+            ("theta_deg", Float64),
+            ("omega_deg_per_day", Float64),
+        ]
+        new_column_names = [n for (n, _) in new_column_names_types]
+        new_column_types = [t for (_, t) in new_column_names_types]
```
cpaniaguam: Why pack these to later unpack them into individual names? Maybe just define new_column_names  with the string literals and attach the types using something like new_column_types = fill(Float64, length(new_column_names)) or a comprehension or explicitly listing the types.

hollandjg: That's a good idea!

src/Tracking/rotation.jl
```diff
@@ -30,6 +30,28 @@ function get_rotation_measurements(
...
+            "theta_deg",
+            "omega_deg_per_day",
+        ]
+        new_column_types = fill(Float64, length(new_column_names))
+        old_column_names_1 = original_names .* "1"
+        old_column_names_2 = original_names .* "2"
+        df = DataFrame(
+            [T[] for T in vcat(new_column_types, original_types, original_types)],
+            vcat(new_column_names, old_column_names_1, old_column_names_2),
+        )
```
cpaniaguam: Consider this slightly cleaner approach using `_add_suffix`.
```suggestion
        measurements_empty = DataFrame(new_names .=> new_column_types)
        df = hcat(measurements_empty, _add_suffix("1", df), _add_suffix("2", df))
```

hollandjg: Good idea, had to modify a bit.

cpaniaguam (approved):

PR #1016, opened 2026-07-01
feat: use NetCDF rather than HDF5 for results archive

hollandjg:
Replace HDF5 with NetCDF support.

Todo:
- [x] drop V1 HDF5 format, remove HDF5 dependency if possible
- [x] make all the masks into integer indexed variables – allow 0 background / 1 cloud, but also add extensibility
- [x] Add creation date
- [x] Add segmentation information as an attribute `ift_configuration` --> load config.txt? or reference to DOI?


src/Archive/V1.jl
```diff
@@ -0,0 +1,533 @@
...
+        defDim(ds, "y", ny)
+        defDim(ds, "time", 1)
+
+        # CRS / geolocation grid_mapping variable (scalar)
+        vcrs = defVar(ds, projection_dataset_name, Int32, ())
+        vcrs.attrib["name"] = crs_name
+        vcrs.attrib["crs_wkt"] = latlondata[:crs_wkt]
+        vcrs.attrib["spatial_ref"] = latlondata[:crs_wkt]
+        vcrs.attrib["long_name"] = "CRS Definition"
+        vcrs.attrib["GeoTransform"] = join(Int64.(latlondata[:geotransform]), " ")
```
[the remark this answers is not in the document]
hollandjg: Changes this to use just the raw float64 representation with scientific form. The file still opened in QGIS, so I think it's okay.

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+
+        # band coordinate variables — string labels describing each channel
+        _band_labels(names, n) =
+            n == length(names) ? names : [names; fill("alpha", n - length(names))]
+        tc_labels = _band_labels(
+            [
+                "R=band_1 (red, 0.620–0.670 µm)",
+                "G=band_4 (green, 0.545–0.565 µm)",
+                "B=band_3 (blue, 0.459–0.479 µm)",
+            ],
```
cpaniaguam: Consider hoisting this list to a global with const. Maybe use a tuple.

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+            nchannels_tc,
+        )
+        fc_labels = _band_labels(
+            [
+                "R=band_7 (mid-IR, 2.105–2.155 µm)",
+                "G=band_2 (NIR, 0.841–0.876 µm)",
+                "B=band_1 (red, 0.620–0.670 µm)",
+            ],
```
cpaniaguam: Also this one.

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+            units="m",
+            standard_name="projection_x_coordinate",
+            long_name="x coordinate of floe centroid",
+        ),
+        "y" => (
+            units="m",
+            standard_name="projection_y_coordinate",
+            long_name="y coordinate of floe centroid",
+        ),
+    )
```
cpaniaguam: Consider extracting this into a helper, e.g. `build_ift_configuration(config, created_at)`, and calling it here. That would centralize the schema and keep this write path focused on I/O.

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+        col_data = props_[!, col_name]
+        T = eltype(col_data)
+
+        v = if T <: Integer
+            defVar(grp, nc_name, Int64, ("floe_label",))
+        elseif T <: AbstractFloat
+            defVar(grp, nc_name, Float64, ("floe_label",); fillvalue=NaN)
+        else
+            defVar(grp, nc_name, String, ("floe_label",))
+        end
```
cpaniaguam: Consider leveraging multiple dispatch for more idiomatic Julian code. E.g.:
```julia
_defVar(grp, name, ::Type{T}, dims) where {T<:Integer}      = defVar(grp, name, Int64,   dims)
_defVar(grp, name, ::Type{T}, dims) where {T<:AbstractFloat} = defVar(grp, name, Float64, dims; fillvalue=NaN)
_defVar(grp, name, ::Type,    dims)                          = defVar(grp, name, String,  dims)
```
Then call it here:
```julia
v = _defVar(grp, nc_name, T, ("floe_label",))
```

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+        # Colour images: stored as (nx, ny, nb) UInt8 → reconstruct Matrix{RGB/RGBA}
+        function read_color(varname)
+            raw = collect(permutedims(Array(ds[varname].var), (3, 2, 1)))  # (nb, ny, nx)
+            nb = size(raw, 1)
+            n0f8 = reinterpret(N0f8, raw)
+            return if nb == 3
+                collect(colorview(RGB{N0f8}, n0f8))
+            else
+                collect(colorview(RGBA{N0f8}, n0f8))
+            end
```
cpaniaguam: A bit cleaner.
```suggestion
            T = nb == 3 ? RGB{N0f8} : RGBA{N0f8}
            return collect(colorview(T, n0f8))
```

src/Archive/V1.jl
```diff
@@ -0,0 +1,538 @@
...
+        4326 => "EPSG:4326 WGS84 lat/lon",
+        3857 => "EPSG:3857 Web Mercator",
+    )
+    crs_name = get(crs_dict, crs_code) do
+        crs_name_ = "EPSG:$(string(crs_code))"
+        @warn "CRS $crs_code not recognized. CRS will be recorded as $crs_name_ in the output file attributes, but no short name will be provided."
+        return crs_name_
+    end
+    return crs_name
+end
```
cpaniaguam:
```suggestion
end

```

cpaniaguam (changes requested):
Thanks for this --  few suggestions.

src/Archive/V1.jl
hollandjg: Yes

src/Archive/V1.jl
hollandjg: Sure!

src/Archive/V1.jl
hollandjg: Sure!

src/Archive/V1.jl
hollandjg: Yes!

src/Archive/V1.jl
hollandjg: Sure

src/Archive/V1.jl
hollandjg: I think an array marked as `const` should be fine.

src/Archive/V1.jl
hollandjg: I think an array marked as `const` should be fine.

src/Archive/V1.jl
hollandjg: I hoisted this to the top of the submodule

cpaniaguam (approved):
Very nice!

PR #1017, opened 2026-07-02
Feat new cloudmask

danielmwatkins:
New cloudmask with simpler methodology and higher coverage of clouds.

1. Candidate cloud <- (B7 > thresh) and (B2 > thresh). This draws a box around the region in the figure, rather than a piecewise linear function.
2. Removing speckle with morphological opening
3. Expand the mask since cloud edges were getting classified as sea ice.
4. Label the connected components in the non-cloud region (holes)
   - If smaller than min floe size: add to mask
   - If contrast is low and the hole is smaller than the max fill size: add to mask

Just using the first step, we cover 90% of the cloud, whereas Rosalinda’s version covered 35% and my update to her method covered 74%. Ice under thin cloud is much more likely to be covered – that’s the tradeoff – at 30%. Still, most thin-cloud sea ice is uncovered.

image

FC images on top, Watkins2025CloudMask in the middle, and Watkins2026CloudMask at the bottom. As you can see, it covers more of the image, and the edges of the mask are smoothed. This is to focus on preventing cloud cover from corrupting the floe retrieval; a more ragged edge may be desirable for other purposes.

I've added in a super-simple test for the new cloud mask and for a couple helper functions.


src/Preprocessing/cloudmask.jl
```diff
@@ -183,7 +186,68 @@ function convert_to_255_matrix(img)::Matrix{Int}
...
+    # remove speckle
+    init_mask .= opening(init_mask, f.opening_strel)
+
+    # expand mask
+    init_mask .= dilate(init_mask, f.dilation_strel)
+
+    # label blank components for object-based methods (background == clouds)
+    labels = label_components(.!init_mask) 
+    remove_small_segments!(labels, f.min_hole_size)
+    maximum(labels) .== 0 && return labels .== 0
```
cpaniaguam, line 242: Scalar to scalar comparison.
```suggestion
    maximum(labels) == 0 && return labels .== 0
```

cpaniaguam: On second thought, I think it's probably ok to just drop this line as `remove_low_contrast_segments!` only touches non-zero labels.

src/Preprocessing/cloudmask.jl
```diff
@@ -233,6 +286,7 @@ function create_cloudmask(
     return f(false_color_image)
 end
 
+# dmw: I think the apply_cloudmask functions can all be removed
```
cpaniaguam, line 289: There is a call to this function in `src/Pipeline/LopezAcosta2019Tiling.jl`.

src/Segmentation/segmented-image-utilities.jl
```diff
@@ -256,3 +256,62 @@ Return an array like `s` where each pixel has the mean intensity or color of its
...
+function remove_small_segments!(labels, min_size)
+    areas = component_lengths(labels)
+    indices = component_indices(labels)
+
+    for L in keys(areas)
+        (L != 0) && begin
+            (areas[L] < min_size) && begin
+                labels[indices[L]] .= 0
+            end
+        end
```
cpaniaguam: `if` with `and` a little cleaner here.
```suggestion
        if L != 0 && areas[L] < min_size
            labels[indices[L]] .= 0
        end
```

src/Segmentation/segmented-image-utilities.jl
```diff
@@ -256,3 +256,62 @@ Return an array like `s` where each pixel has the mean intensity or color of its
...
+    indices = component_indices(labels)
+
+    for L in keys(areas)
+        (L != 0) && begin
+            (areas[L] > max_size) && begin
+                labels[indices[L]] .= 0
+            end
+        end
+    end
+end
```
cpaniaguam: This function and its counterpart only differ in the comparison. You could do something like this:
```julia
function _remove_segments!(labels, op, threshold)
    areas = component_lengths(labels)
    indices = component_indices(labels)
    for L in keys(areas)
        if L != 0 && op(areas[L], threshold)
            labels[indices[L]] .= 0
        end
    end
end

remove_small_segments!(labels, min_size) = _remove_segments!(labels, <, min_size)
remove_large_segments!(labels, max_size) = _remove_segments!(labels, >, max_size)

File: src/Segmentation/segmented-image-utilities.jl [hunk 5]
```diff
@@ -256,3 +256,62 @@ Return an array like `s` where each pixel has the mean intensity or color of its
...
+    indices = component_indices(labels)
+    for L in unique(labels)
+        (L != 0) && (areas[L] < max_fill_size) && begin
+            contrast = maximum(ref_img[indices[L]]) - minimum(ref_img[indices[L]])
+            (contrast < min_contrast) && begin
+                labels[indices[L]] .= 0
+            end
+        end
+    end
+end
```
cpaniaguam, line 308: This one _feels_ kind of similar too but I guess it's tricker to have them all share the loop with the comparison.

cpaniaguam (approved):
I like this! Left a couple optional suggestions.

src/Preprocessing/cloudmask.jl
danielmwatkins: I'll test that out, thanks for thinking through this!

src/Preprocessing/cloudmask.jl
danielmwatkins: Right -- I don't think we can just delete them without care, what I think is that the core function should be covered by "apply mask" and the specialized use, where it's apply mask and then erase one band, is only used in one workflow (if I recall correctly). So, that specialized use should really be a helper function within that pipeline, and everything else can just use a universal masker function. Only after that is taken care of can we remove the apply cloud mask functions.

src/Segmentation/segmented-image-utilities.jl
danielmwatkins: Agreed

src/Segmentation/segmented-image-utilities.jl
danielmwatkins: That's cool, I always forget about how we can pass around math symbols as functions.

src/Segmentation/segmented-image-utilities.jl
danielmwatkins: Right, the operation isn't quite as easy to send in. Perhaps the real generalized idea is to have an operation which is the criteria for removing a segment, and a function is passed through to specify the criteria. It'll take more thought to figure out the right approach though.

cpaniaguam (approved):
lgtm! Opened https://github.com/WilhelmusLab/IceFloeTracker.jl/issues/1019 to address the test dataset downloding flakiness in CI at some point.

PR #1018, opened 2026-07-03
Clean up code by removing duplicated ice detection algorithm and deprecated format for kmeans binarization

danielmwatkins:
IceDetectionLopezAcosta2019 was defined in two places -- the pipeline and in the Segmentation file. ice-labels duplicated code in find-ice-masks.


src/Pipeline/Pipeline.jl
```diff
@@ -0,0 +1,7 @@
+module Pipeline
+
+export IceFloeSegmentationAlgorithm
+
+include("abstract-algorithms.jl")
+
+end
```
cpaniaguam:
```suggestion
end

```

cpaniaguam (approved):
Lgtm! One little thing to consider.

cpaniaguam (approved):

PR #1022, opened 2026-07-10
workflow: include configuration with large regions

hollandjg:
... for easy testing of performance on large regions.


danielmwatkins (approved):
This is great, thanks for adding this!

PR #1026, opened 2026-07-17
Refactor default filter setup in FilterFunction

cpaniaguam:


danielmwatkins (commented):
This approach makes sense to me. Question though, what type is the default list going to have? Will making that a const be a limitation?

danielmwatkins (approved):
I like it! Thanks for the explanation.

